PR #11, opened 2023-04-14
WIP Document data structures as UML

Gweaton:
This is still early stages and to be revised, but we've been trying to use schema.org terminology as much as possible to keep things flexible.

## WIP diagram

image


Gweaton:
Closing this in favour of the [documentation in the API](https://github.com/ministryofjustice/hmpps-accredited-programmes-api/pull/14) - we may want to add our own documentation here to explain how Programmes are mapped to Courses etc once we start building things in the UI.

PR #14, opened 2023-04-18
Add `.env.example`

yndajas:
## Context

This is used in the setup script to copy into `.env`, but it was missing

## Changes in this PR

Add missing file used in setup script


Gweaton (approved):
👍

PR #17, opened 2023-04-21
(355) Refactor routes and replace index with dashboard

Gweaton:
## Context

https://trello.com/c/3knwWIrc/355-build-refactor-dashboard-page

## Changes in this PR

This PR adds a new DashboardController which is responsible for
rendering the page users see when they log into the service. The "Find
an accredited programme" link doesn't currently go anywhere, but will
point to the new "Find" journey, which we'll be building very soon.

Controllers will be passed to the `routes()` function directly to allow
us to use the controller actions directly when mapping them to a
specific route. This has required some re-jigging further up the chain,
via the new `controllers/index.ts` file that will serve as a "hub" of
sorts for all the controllers in the application that can then be
injected into the running application.

### Before

image

### After

image


server/views/dashboard/index.njk
```diff
@@ -5,7 +5,7 @@
 
 {% block content %}
 
-  <h1>This site is under construction...</h1>
-  <p>Please check back later when there is content to view.</p>
+  <h1>{{ pageHeading }}</h1>
 
+  <a class="govuk-link" href="#">Find an accredited programme</a>
```
yndajas, line 10: Is there a reason for paremeterising the heading but not the link text?

server/errorHandler.test.ts
```diff
@@ -1,20 +1,14 @@
...
+import DashboardController from './controllers/dashboardController'
 
 afterEach(() => {
   jest.resetAllMocks()
 })
 
 describe('GET 404', () => {
   it('should render content with stack in dev mode', () => {
-    return request(app)
+    return request(appWithAllRoutes({ controllers: { dashboardController: new DashboardController() } }))
```
yndajas, line 11: Isn't the object here and below duplicating code in `server/controllers/index.ts`? Can't we just import from that instead of importing the `DashboardController` and recreating it?

server/routes/index.ts
```diff
@@ -1,16 +1,16 @@
 import { type RequestHandler, Router } from 'express'
 
 import asyncMiddleware from '../middleware/asyncMiddleware'
-import type { Services } from '../services'
+import { Controllers } from '../controllers'
```
yndajas, line 4: Is there a reason for removing `type`? Approved Premises still has it

yndajas (approved):
A few queries we can go over at another time, but approving and will merge for now while you're off

server/views/dashboard/index.njk
yndajas: I think the question is more "Is there a reason for parameterising page headings when they're static?"?

Gweaton: I think the main reason we parameterise on this project is so we can reuse the `pageHeading` variable for the `pageTitle` above on line 3 (which should be `applicationName + " - " + pageHeading` rather than `applicationName + "- Home`"). We're not currently doing this, but we should be!

Otherwise, there's not a whole lot of advantage to it, other than consistency across the application if we were to have any titles that'd include dynamic content - one use I can imagine would be if we were translating the site we could possibly pass in some translations there, but we might also nunjucks view helpers for that.

yndajas: > We're not currently doing this, but we should be!

Should we? I assumed we keep them separate because we don't want the page title to be as long as the main heading might usefully be?

Yeah, it would help for translations, but then we should do the same for all the text content 😅

I suppose it does keep things consistent regardless of whether it's a static or dynamic/interpolated heading, although arguably it just shifts that distinction to the controller

Gweaton: I don't think we need to worry about headings getting all that long to be honest - if they do we can handle it on a case-by-case basis. It looks like [this is the approach](https://github.com/search?q=repo%3Aministryofjustice%2Fhmpps-approved-premises-ui%20pageHeading&type=code) they've taken in the Approved Premises project too

PR #18, opened 2023-04-21
(330) Run instance of API locally & add basic /programmes index page

Gweaton:
## Context

https://trello.com/c/Fp5srGJw/329-add-script-for-running-local-api-against-ui
https://trello.com/c/v5PcT05d/330-configure-api-client-in-the-ui

## Changes in this PR

- Starts up the Accredited Programmes API locally as a docker container as part of the script for starting backing services.
- Adds a "/programmes" page for rendering the Programmes provided by the API's `/programmes` endpoint.


yndajas (approved):

server/controllers/dashboardController.ts
```diff
@@ -1,10 +1,12 @@
 import { Request, RequestHandler, Response } from 'express'
+import paths from '../paths/api'
```
Gweaton, line 2: I think we should create a new `paths/find.ts` file for this - we're using the API's `/programmes` path here which will change to `/courses` in the future, but we want to keep it as `/programmes` in the client.

Gweaton (commented):
Just spotted one last thing, otherwise we're good to go.

PR #21, opened 2023-05-04
(289) Build programmes list

yndajas:
## Context

#18 connects with the API, creates a basic programme list and handles a bunch of test setup etc. This originally included pagination, but this will need more work on the API side to make it functional - there's little point building it out and adding tests until we know how it will work

## Changes in this PR

This tidies up the programme list, using the information provided by the API where possible, and hardcoded information where it's yet to come

### Before

image

### After

image


yndajas:
[Build programme list (UI)](https://trello.com/c/WvncTFdV/289-build-programme-list-ui)

integration_tests/pages/find/programmes.ts
```diff
@@ -1,14 +1,26 @@
...
   }
 
   shouldHaveProgrammes(programmes: Array<AccreditedProgramme>) {
-    cy.get('li').each((programmeElement, programmeElementIndex) => {
-      cy.wrap(programmeElement).should('have.text', programmes[programmeElementIndex].name)
+    cy.get('div[role=listitem]').each((programmeElement, programmeElementIndex) => {
+      cy.wrap(programmeElement).within(() => {
+        const programme = programmes[programmeElementIndex]
+        cy.get('h2').should('have.text', programme.name)
+        // TODO: test for tags/audience when implemented in API
```
Gweaton, line 15: Can we test for these now, even if they're hard coded?

server/utils/programmeUtils.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  describe('programmeListItems', () => {
+    it('converts programmes into programme list items (converting the prerequisites)', () => {
+      const programmes = programmeFactory.buildList(3)
+
+      const expected = programmes.map(programme => {
+        return {
+          ...programme,
+          programmePrerequisites: prerequisiteSummaryListRows(programme.programmePrerequisites),
+        }
+      })
```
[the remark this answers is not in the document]
Gweaton: Yeah, I think it would be good to write this all out - we'll only need to do it once as this is a util that'll be reused, but I think it would also help the tests serve as standalone documentation of the code, rather than tests of functionality.

server/utils/programmeUtils.test.ts
```diff
@@ -0,0 +1,31 @@
+import { programmeFactory, programmePrerequisiteFactory } from '../testutils/factories'
+import { prerequisiteSummaryListRows, programmeListItems } from './programmeUtils'
+
+describe('programmeUtils', () => {
+  describe('prerequisiteSummaryListRows', () => {
+    it('converts prerequisites into summary list rows', () => {
+      const prerequisites = programmePrerequisiteFactory.buildList(3)
+
+      const expected = prerequisites.map(prerequisite => {
+        return { key: { text: prerequisite.key }, value: { text: prerequisite.value } }
```
Gweaton, line 10: Can we do the text formatting/capitalisation here? It _could_ be the responsibility of the API, but I think it's safest to do this in the UI so we can be sure it'll always look consistent.

Gweaton (commented):
Looks good, just a couple of small things

server/utils/programmeUtils.test.ts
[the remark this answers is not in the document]
yndajas: I've updated this - what do you think about the `programmeListItems` format? It's slightly more explicit, but I still kept the spreading and the use of `prerequisiteSummaryListRows`. The spreading indicates "it returns a value that has all the key/value pairs of the original programmes", then the use of `prerequisiteSummaryListRows` indicates "except that the value of `programmePrerequisites` is overwritten with the output of this method". Since that method is already tested above, it feels like we shouldn't be testing it's output here too, only that this method overwrites `programmePrerequisites` with its output?

server/utils/programmeUtils.test.ts
yndajas: We could, but then would we not also want to do that with every text value that comes from the API, e.g. the title, description etc, if we can't rely on the API giving us clean data?

Gweaton: That's a good point. I suppose as we're in control of the API we can just ensure it comes back in the right format. When using other APIs we may need to revisit this, but I think I'm happy to sort this on the API side.

server/@types/ui/index.d.ts
```diff
@@ -0,0 +1,14 @@
...
+  key: {
+    text: string
+  }
+  value: {
+    text: string
+  }
+}
+
+export type ProgrammeListItem = Omit<AccreditedProgramme, 'programmePrerequisites'> & {
+  programmePrerequisites: Array<SummaryListRow>
```
Gweaton, line 13: Sorry, I missed this on my original review. To avoid confusion in the nunjucks templates and to make this extra clear, could we call this key `programmePrerequisitesSummaryListRows` so it's clear it's some nunjucks-formatted objects?

server/utils/programmeUtils.test.ts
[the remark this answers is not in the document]
Gweaton: Yeah I think this is fine - it's pretty clear from reading the test what's going on now, thanks for updating.

Gweaton (approved):
Looks good, good to merge after updating the last commit with an explanation of why Lodash is necessary.

PR #22, opened 2023-05-04
Sort imports

yndajas:
## Context

On the Approved Premises UI repo, a linting rule dictates that when importing multiple things from one module, the imports are alphabetised. This like a useful thing to do here too

## Changes in this PR

This adds a linter check for ordering imports alphabetically (within a single line, not between lines) and fixes resulting errors


Gweaton (approved):
Nice

PR #25, opened 2023-05-09
Add djLint

yndajas:
## Context

We currently have ESLint and Prettier configured to check TypeScript, JavaScript, CSS and JSON files, but nothing for Nunjucks files

## Changes in this PR

This adds djLint, which appears to be the most popular option for linting/formatting Nunjucks files. That said, it doesn't appear to have a bunch of discussion around it online, and some of the formatting changes are a little questionable. It seems a little less mature at formatting than solutions like ESLint, Prettier and Standard, but I think that **some kind of consistent automated formatting is better than manual/inconsistent formatting**. It's manipulatable to a degree, so you can adjust the results of it's automatic format, run it again and it won't _always_ change it back (so it has a range of acceptable answers, some of which make more sense)

Might make sense to squash at least the first two commits, to avoid a red commit


integration_tests/pages/find/programmes.ts
```diff
@@ -8,7 +8,10 @@ export default class ProgrammesPage extends Page {
 
   shouldHaveProgrammes(programmes: Array<AccreditedProgramme>) {
     cy.get('li').each((programmeElement, programmeElementIndex) => {
-      cy.wrap(programmeElement).should('have.text', programmes[programmeElementIndex].name)
+      cy.wrap(programmeElement)
```
Gweaton, line 11: Do we think it's likely that we'll need to do this all over the place if the preferred style is to the content of an element on its own line? Is it worth us extracting this into its own Cypress command if so?

PR #27, opened 2023-05-10
Add linting and formatting

yndajas:
## Context

#25 (unmerged) added djLint for both linting and formatting of Nunjucks files. The formatting side had some very questionable choices. After consulting with the Approved Premises team, I've tested another approach, which would replace #25

## Changes in this PR

This uses Nunjucks Template Formatter (NTF) for formatting, which is just a VSCode extension. Not ideal if we have team members who don't use VSCode, but it does formats templates much more sensibly. For formatting and for our current team setup, this is probably fine

djLint is still used for linting, including in `npm run lint`, used in `script/test` (but skipped in the CI). One rule is turned off due to conflicting with a formatting choice of NTF that isn't immediately obviously configurable

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,7 +49,7 @@ jobs:
             npm run record-build-info
       - run: # Run linter after build because the integration test code depend on compiled typescript...
           name: Linter check
-          command: npm run lint
+          command: npm run lint:ci
       - persist_to_workspace:
           root: .
           paths:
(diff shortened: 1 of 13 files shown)


Gweaton (approved):
I think this is sensible, and I prefer the way the NTF formats things - thanks for sticking with it. Does djlint's linting not complain that NTF has formatted things in a different way than it would have done?

PR #37, opened 2023-05-15
Sort imports redux

yndajas:
## Context

#22 added sorting of members within multiple member import declarations, but not sorting between import declarations

## Changes in this PR

- Add sorting and grouping between import lines
- Simplify/tidy up ESLint config
- Make Cypress folders more consistent

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,86 +5,41 @@
     "jest": true
   },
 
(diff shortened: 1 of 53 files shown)


Gweaton (approved):
Nice, this looks a lot neater and glad we'll get some linting on the imports too

PR #41, opened 2023-05-17
Update backing service images in `script/update`

yndajas:
## Context

We currently have to remember to run `docker compose pull` to bring in API changes

## Changes in this PR

This updates all backing service images when running `script/update`. This is automatically run in `script/test`, but an option to run `script/test` without the update (to images or any other dependencies) is provided


Gweaton (approved):
Nice one

PR #44, opened 2023-05-18
(403) Add Prison client

Gweaton:
## Context

https://trello.com/c/d4yEqtSO/403-fetch-prison-information

## Changes in this PR

This adds a Prison client for fetching Prison data from the Prisons API. We'll use it to populate the "Offerings" page in the near future. It also starts up the Prisons API locally as a backing service.

There's still an ongoing conversation as to whether this is the correct API to use, but it does seem to have what we need, for now at least. We'll wrangle the Prison object we get back into our own internal `Organisation` model in future work.

### DIFF
--- a/server/data/prisonClient.ts
+++ b/server/data/prisonClient.ts
@@ -1,0 +1,17 @@
+import RestClient from './restClient'
+import type { ApiConfig } from '../config'
+import config from '../config'
+import paths from '../paths/prisonApi'
+import type { Prison } from '@prison-api'
+
+export default class PrisonClient {
+  restClient: RestClient
+
+  constructor(token: string) {
+    this.restClient = new RestClient('prisonClient', config.apis.prisonApi as ApiConfig, token)
+  }
+
+  async getPrison(agencyId: string): Promise<Prison> {
+    return (await this.restClient.get({ path: paths.prisons.show({ agencyId }) })) as Prison
+  }
+}
--- a/server/testutils/factories/prison.ts
+++ b/server/testutils/factories/prison.ts
@@ -1,0 +1,24 @@
+import { faker } from '@faker-js/faker/locale/en_GB'
+import { Factory } from 'fishery'
+
+import prisonAddressFactory from './prisonAddress'
+import type { Prison } from '@prison-api'
+
+export default Factory.define<Prison>(({ params }) => {
+  const locality = params.locality || faker.location.county()
+  const city = params.city || faker.location.city()
+  const country = params.country || 'United Kingdom'
+  const premise = params.premise || `HMP ${locality}`
+
+  return {
+    agencyId: faker.string.alpha({ length: 3, casing: 'upper' }),
+    premise,
+    locality,
+    city,
+    country,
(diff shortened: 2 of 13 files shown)


server/data/prisonClient.ts
```diff
@@ -0,0 +1,16 @@
+import RestClient from './restClient'
+import { Prison } from '../@types/prisonApi'
+import config, { ApiConfig } from '../config'
+import paths from '../paths/prisonApi'
+
+export default class PrisonClient {
+  restClient: RestClient
+
+  constructor(token: string) {
+    this.restClient = new RestClient('programmeClient', config.apis.prisonApi as ApiConfig, token)
```
yndajas, line 10: `'prisonClient'`

yndajas (commented):
Looks good. Just a few small things. Also, should the prisons API be accessible on 9091 when running `script/server`? I'm not seeing it

server/data/prisonClient.ts
Gweaton: oops!

server/testutils/factories/prison.ts
```diff
@@ -0,0 +1,31 @@
...
+      {
+        premise,
+        street: streetAddress,
+        locality: county,
+        town: city,
+        postalCode: faker.location.zipCode(),
+        country,
+        primary: true,
+      },
+    ],
```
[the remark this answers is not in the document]
Gweaton: I've had a go a this - let me know what you think

yndajas (commented):
A few more queries and minor things to clean up!

server/testutils/factories/prisonAddress.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  return {
+    premise,
+    street: faker.location.streetAddress(),
+    locality,
+    town: faker.location.city(),
+    postalCode: faker.location.zipCode(),
+    country: faker.location.country(),
+    primary: true,
+  }
```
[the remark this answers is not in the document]
Gweaton: Fishery does this for us, so it uses the value pass in as a parameter, or if one isn't passed in, it uses the value you've defined in the factory.

server/testutils/factories/prisonAddress.ts
```diff
@@ -0,0 +1,19 @@
+import { faker } from '@faker-js/faker'
```
yndajas, line 1: We can use `@faker-js/faker/locale/en_GB` to get proper UK postcodes (and more UK-style data in general). Can do the same in `prison`, but the post code is where it most noticeably affects the format

yndajas (approved):
Looks good. Just one minor thing that might be good to update then I think it's ready to go

server/testutils/factories/prisonAddress.ts
Gweaton: Good spot, have added this and hardcoded 'United Kingdom' as it didn't make sense for things like "Uganda" to show with a UK postcode. We can revisit this in future if we need but I don't think we'll need to go more granular in our test data just yet.

PR #46, opened 2023-05-18
Sort `package.json` scripts

yndajas:
## Context

Currently a lot of the `package.json` scripts reference each other and they're also referenced in some of our scripts to rule them all. They're not in a very clear order, and often you have to search around to find one that's referenced in another one. The list is getting quite long, which makes this harder

## Changes in this PR

This alphabetises the list to make it easy to find what you're looking for


Gweaton (approved):
Thanks for this!

PR #47, opened 2023-05-24
(362) Pass user token with request to API

Gweaton:
## Context

[Trello card](https://trello.com/c/UwKlC3n8/362-add-token-to-http-request-for-authenticating-with-our-api-after-hmpps-auth-has-been-added)

## Changes in this PR

1. Sets the `HMPPS_AUTH_URL` environment variable on the local Accredited Programmes API run with Docker, as it's missing on start up with our local configuration now that the API's been secured behind HMPPS Auth.
2. We now pass the logged-in-user's auth token with our request to the Accredited Programmes API to authenticate with it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,6 +21,8 @@ services:
       - "9092:8080"
     healthcheck:
       test: ['CMD', 'curl', '-f', 'http://localhost:8080/health']
+    environment:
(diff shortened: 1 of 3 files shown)


yndajas (approved):
Seems to be working for me locally (after running `script/update`)

PR #49, opened 2023-05-26
(359) Add course show page

yndajas:
## Changes in this PR

This primarily adds the course show page, AKA the programme and offerings page. There are a few smaller bits in here:
- fixing Wiremock scripts
- making a util method more generic (to reuse in the show page)
- fixing a page heading
- renaming a template macro for clarity
- using predefined path when stubbing API endpoints

## Screenshots of UI changes

image


wiremock/scripts/stubApis.ts
```diff
@@ -8,7 +8,7 @@ stubs.push(async () =>
   stubFor({
     request: {
       method: 'GET',
-      url: '/programmes',
+      url: '/courses',
```
yndajas, line 11: Why don't tests fail without this?

yndajas, line 11: Use path imported from `paths/api` here and elsewhere in the file

package.json
```diff
@@ -131,6 +131,7 @@
     "@types/express-session": "^1.17.7",
     "@types/http-errors": "^2.0.1",
     "@types/jest": "^29.5.0",
+    "@types/jest-when": "^3.5.2",
```
yndajas, line 134: Separate commit (all `package.json` changes)

server/controllers/find/coursesController.test.ts
```diff
@@ -42,4 +46,38 @@ describe('CoursesController', () => {
       expect(courseService.getCourses).toHaveBeenCalledWith(token)
     })
   })
+
+  describe('show', () => {
+    it('show render the course show template', async () => {
+      const course = courseFactory.build()
+      courseService.getCourse.mockResolvedValue(course)
+
+      const courseOfferings = courseOfferingFactory.buildList(10)
```
yndajas, line 55: Make this 3 to reduce the looping given we aren't testing anything related to the number

server/controllers/find/coursesController.test.ts
```diff
@@ -49,26 +49,23 @@ describe('CoursesController', () => {
...
-          const organisation = organisationFactory.build({ id: organisationId })
-          organisations.push(organisation)
-          when(organisationService.getOrganisation).calledWith(token, organisationId).mockResolvedValue(organisation)
+        const courseOfferings = courseOfferingFactory.buildList(3)
+        courseService.getOfferingsByCourse.mockResolvedValue(courseOfferings)
+
+        const organisationIds = courseOfferings.map(offering => offering.organisationId)
+        const organisations = courseOfferings.map(offering =>
+          organisationFactory.build({ id: offering.organisationId }),
+        )
```
yndajas, line 66: We can just map over the IDs from the last line here

server/controllers/find/coursesController.test.ts
```diff
@@ -89,13 +86,24 @@ describe('CoursesController', () => {
 
     describe('when some but not all organisations are not found by the organisation service', () => {
```
yndajas, line 87: `when some but not all organisations are --not-- found by the organisation service`

server/controllers/find/coursesController.test.ts
```diff
@@ -89,13 +86,24 @@ describe('CoursesController', () => {
...
-        organisationIds.forEach((organisationId, organisationIdIndex) => {
-          if (organisationIdIndex) {
-            continue
-          }
-          const organisation = organisationFactory.build({ id: organisationId })
-          organisations.push(organisation)
-          when(organisationService.getOrganisation).calledWith(token, organisationId).mockResolvedValue(organisation)
+        const courseOfferings = courseOfferingFactory.buildList(2)
+
+        const courseOfferingsWithUnfindableCourse = [
```
yndajas, line 91: `UnfindableCourse` -> `UnfindableOrganisation`

server/controllers/find/coursesController.test.ts
```diff
@@ -89,13 +86,24 @@ describe('CoursesController', () => {
...
+          courseOfferingFactory.build({ organisationId: 'NOTFOUND' }),
+        ]
+
+        courseService.getOfferingsByCourse.mockResolvedValue(courseOfferingsWithUnfindableCourse)
+
+        const organisationIds = courseOfferingsWithUnfindableCourse.map(offering => offering.organisationId)
+
+        const organisations = courseOfferingsWithUnfindableCourse.map(offering =>
+          organisationFactory.build({ id: offering.organisationId }),
+        )
```
yndajas, line 102: Same with the IDs here - but do we even need to get the IDs? We could just do this and remove the IDs map

server/controllers/find/coursesController.test.ts
```diff
@@ -106,10 +114,14 @@ describe('CoursesController', () => {
           expect(organisationService.getOrganisation).toHaveBeenCalledWith(token, organisationId),
         )
 
+        const organisationsWithoutUnfindableCourse = organisations.filter(
+          organisation => organisation.id !== 'NOTFOUND',
+        )
```
yndajas, line 119: I think we could also reduce some of the looping by creating this variable higher up and then creating an extra variable with the unfindable course?

server/data/courseClient.ts
```diff
@@ -14,4 +14,12 @@ export default class CourseClient {
   async all(): Promise<Array<Course>> {
     return (await this.restClient.get({ path: paths.courses.index({}) })) as Array<Course>
   }
+
+  async find(id: Course['id']): Promise<Course> {
+    return (await this.restClient.get({ path: paths.courses.show({ id }) })) as Course
+  }
+
+  async offerings(courseId: Course['id']): Promise<Array<CourseOffering>> {
```
Gweaton, line 22: I think we should call this `findCourseOfferings` or something similar to keep it consistent with our "verby" function name convention in our clients.

server/views/courses/_locationsTable.njk
```diff
@@ -0,0 +1,31 @@
...
+      }
+    },
+    {
+      text: "Region",
+      attributes: {
+        "aria-sort": "none"
+      }
+    },
+    {
+      text: ""
```
Gweaton, line 27: Did we want to give this an `Action` heading or anything similar?

Gweaton (approved):
Looking good - just a couple of tiny things before merging.

server/views/courses/_locationsTable.njk
yndajas: Oh yeah, I asked about that in a couple of channels and forgot to update it! What's your preference out of:
- keeping it visually like the design, but with a `govuk-visually-hidden` heading
- adding "Action", against the design?

Also need to add more visually hidden information on each of the links, e.g. the prison name

server/data/courseClient.ts
Gweaton: Yeah, I don't think it's a huge deal but I think it'd be good to aim for the verby style in both

server/views/courses/_locationsTable.njk
Gweaton: I'd probably opt for the latter, but maybe one to ask Chanelle?

server/data/courseClient.ts
yndajas: So we have the following client/service mappings:
- `getAll` -> `getCourses`
- `find` -> `getCourse`
- `getOfferings` -> `getOfferingsByCourse`

Is it worth bringing them fully in sync, following the service method naming convention? I guess `all` and `find` are Active Record like conventions, but if we want the clients to be verby like the services, should we go all the way with the consistency?

Gweaton: I like how you've mapped them out there in your comment - can we do that?

yndajas: Yep, that's how I have it at the moment (locally), just wondering how we conceptualise the distinction

yndajas: Hmm, but if it was more Rails-y, offerings would probably be an association, in which case it would just be `course.offerings`, but then this is a client not a model 🤔

`find` is typically for a single record in Active Record, but `where` or `select` don't feel appropriate either

Maybe `findOfferings` is fine for now

PR #55, opened 2023-06-01
Fix PR template styling

yndajas:
## Context

This checklist item is appearing with line breaks as per the code when in a PR description (but not if you just view the template on GitHub), which is not what we want

## Changes in this PR

Though it's best to keep lines short, we'll make an exception here to make it show up properly on PR descriptions

### Before

image

### After

image

## Post-merge checklist

- [ ] Have you written an end-to-end test for the happy path in the [Accredited Programmes E2E repo](https://github.com/ministryofjustice/hmpps-accredited-programmes-e2e) (if necessary)?

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -14,7 +14,4 @@
 
 ## Pre merge checklist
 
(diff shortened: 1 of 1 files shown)


Gweaton (approved):
thank you!

PR #57, opened 2023-06-01
Use a specific course ID in Pact contract tests

Gweaton:
## Changes in this PR

We were previously using a randomly generated ID provided by the factory, meaning the provider would have to generate exactly the same ID to fulfil these contracts and make the tests pass.

Here we hardcode the course ID so we can use the same one in the Provider's contract verification tests.

### DIFF
--- a/server/data/courseClient.test.ts
+++ b/server/data/courseClient.test.ts
@@ -17,11 +17,15 @@ pactWith({ consumer: 'Accredited Programmes UI', provider: 'Accredited Programme
   })
 
   const allCourses = courseFactory.buildList(3)
+  const course = courseFactory.build({
+    id: '28e47d30-30bf-4dab-a8eb-9fda3f6400e8',
+  })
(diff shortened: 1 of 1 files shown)


server/data/courseClient.test.ts
```diff
@@ -17,11 +17,15 @@ pactWith({ consumer: 'Accredited Programmes UI', provider: 'Accredited Programme
   })
 
   const allCourses = courseFactory.buildList(3)
+  const course = courseFactory.build({
+    id: '28e47d30-30bf-4dab-a8eb-9fda3f6400e8',
+  })
```
yndajas, line 22: Do we need to make the ID in the course factory `params.id || ` for this to work (we do that for some other fields)?

yndajas: It appears not, and it looks like we can remove this from the `course` and `courseAudience` factories (I think we said we only need it when the `||` statement is in a `const`

yndajas (approved):

PR #59, opened 2023-06-02
Remove explicit factory params

yndajas:
## Context

We're explicitly using params here in a way that's built in with Fishery, so it's redundant: we get this behaviour for free

## Changes in this PR

Default to implicit param overrides in factories


Gweaton (approved):
Ah yep - sorry I must've missed this when reviewing

PR #65, opened 2023-06-12
(497) Use paths in routes

yndajas:
## Context

We were still hardcoding paths in our router. Per Approved Premises, we can get the path patterns with the `.pattern` method. Some of the groundwork for this in Approved Premises was done in https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/138, which adds namespacing in controllers, paths and routers. We already have this in our paths, but not in our controllers or routers

Also related to paths, we import a `paths` object from three different sources across the app, which can be a little confusing when reading code snippets

## Changes in this PR

- clarify which paths are being used where
- namespace controllers and routers
- use path patterns in routes

### DIFF
--- a/server/utils/routeUtils.ts
+++ b/server/utils/routeUtils.ts
@@ -1,0 +1,9 @@
+import type { RequestHandler, Router } from 'express'
+
+import asyncMiddleware from '../middleware/asyncMiddleware'
+
+export default function actions(router: Router) {
+  return {
+    get: (path: string | Array<string>, handler: RequestHandler) => router.get(path, asyncMiddleware(handler)),
+  }
+}
--- a/integration_tests/e2e/find.cy.ts
+++ b/integration_tests/e2e/find.cy.ts
@@ -1,4 +1,4 @@
-import paths from '../../server/paths/find'
+import findPaths from '../../server/paths/find'
 import { courseFactory, courseOfferingFactory, prisonFactory } from '../../server/testutils/factories'
 import { organisationFromPrison } from '../../server/utils/organisationUtils'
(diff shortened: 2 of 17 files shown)


server/utils/routeUtils.ts
```diff
@@ -0,0 +1,9 @@
+import type { RequestHandler, Router } from 'express'
+
+import asyncMiddleware from '../middleware/asyncMiddleware'
+
+export default function actions(router: Router) {
+  return {
+    get: (path: string | Array<string>, handler: RequestHandler) => router.get(path, asyncMiddleware(handler)),
+  }
```
yndajas, line 8: In Approved Premises this was in `server/routes/utils.ts` and didn't have an accompanying unit test. I put it in the main `utils` folder like we've done with `typeUtils`. The method in `typeUtils` in our repo doesn't have a unit test either, but I added a card for doing this (if we want it). Shall we add unit tests for both?

Gweaton: Yeah, it'd be good to have a unit test for this in this PR, and we can add another one for that other method when we pick up that technical debt card.

Gweaton (approved):
Looks good, approving now but it'd be good to add that unit test.

yndajas:
[Use `paths` with `.pattern` in `routes/index` (UI)](https://trello.com/c/TsioFWgW/497-use-paths-with-pattern-in-routes-index-ui)

server/utils/routeUtils.ts
yndajas: To be completed separately, since the mocking is a little complicated

PR #66, opened 2023-06-12
Add test sentry error route

Gweaton:
## Changes in this PR

This is a temporary route that will throw an error so we can confirm our sentry config is working as expected.


yndajas (approved):
Congratulations!

PR #70, opened 2023-06-15
(491) Switch to Prison Register API

yndajas:
## Context

We've been advised to switch to the Prison Register API to retrieve information about prisons

## Changes in this PR

Switch to the Prison Register API, plus a couple of minor bonuses:
- use a single index file for prison-related types
- add a `NonEmptyArray` type


server/testutils/factories/prison.ts
```diff
@@ -2,23 +2,26 @@ import { faker } from '@faker-js/faker/locale/en_GB'
...
-import type { Prison } from '@prison-api'
+import type { Prison } from '@prison-register-api'
 
 export default Factory.define<Prison>(({ params }) => {
-  const locality = params.locality || faker.location.county()
-  const city = params.city || faker.location.city()
-  const country = params.country || 'United Kingdom'
-  const premise = params.premise || `HMP ${locality}`
+  const county = faker.location.county()
+  const prisonName = params.prisonName || `${county} (HMP)`
```
yndajas, line 9: Probably fine as a factory default, but if we're supporting YOIs as well, we could randomise this with the `arrayElement` helper again

server/testutils/factories/prison.ts
```diff
@@ -2,23 +2,26 @@ import { faker } from '@faker-js/faker/locale/en_GB'
...
+    female: faker.datatype.boolean(),
+    contracted: faker.datatype.boolean(),
+    types: [
+      {
+        code: 'HMP',
+        description: "Her Majesty's Prison",
+      },
     ],
+    addresses: [prisonAddressFactory.build({ county })],
+    operators: [{ name: faker.helpers.arrayElement(['G4S', 'MTCNovo', 'PSP', 'Serco', 'Sodexo']) }],
```
yndajas, line 25: Would it be better to use some fake provider names for this (even though the dev data these are from is public)?

server/testutils/factories/organisationAddress.ts
```diff
@@ -4,9 +4,10 @@ import { Factory } from 'fishery'
 import type { OrganisationAddress } from '@accredited-programmes/models'
 
 export default Factory.define<OrganisationAddress>(() => ({
-  addressLine1: faker.location.streetAddress(),
+  addressLine1: faker.helpers.arrayElement([null, faker.location.streetAddress()]),
```
Gweaton, line 7: How likely is it that these address fields could be empty based on the `dev` data you've seen?

I wonder if for our general factory here we should assume the best, i.e. that these fields are present, and only set `null` values in specific test cases where we want to handle this null behaviour. Otherwise, any time we want to have a full address we'll need to override these fields.

server/testutils/factories/prison.ts
Gweaton: I think this is fine - if we do want to add YOIs we can set them manually.

server/testutils/factories/prison.ts
Gweaton: I don't think it's too much of an issue for the reason you've specified, but let's make these a bit more fake to be safe.

server/testutils/factories/prisonAddress.ts
```diff
@@ -1,19 +1,14 @@
...
-    locality,
-    town: faker.location.city(),
-    postalCode: faker.location.zipCode(),
-    country: 'United Kingdom',
-    primary: true,
-  }
-})
+export default Factory.define<PrisonAddress>(() => ({
+  id: faker.number.int({ min: 1 }),
+  addressLine1: faker.helpers.arrayElement([null, faker.location.streetAddress()]),
```
Gweaton, line 8: My message about addresses above applies here too - let's remove the `null`ness from these factories apart from when setting `addressLine2` as that feels like a more real-world scenario.

server/utils/organisationUtils.test.ts
```diff
@@ -10,116 +12,174 @@ import {
...
         id: 'an-ID',
-        name: prison.premise,
+        name: prison.prisonName,
         category: 'N/A',
         address: {
           addressLine1: 'Rabbit Avenue',
+          addressLine2: null,
           town: 'Leporidaeville',
-          county: 'Not found',
+          county: null,
```
Gweaton, line 42: Shall we make this a real county?

Gweaton (commented):
Nice work - just a couple of comments re: factory data that we should address, but otherwise this approach looks good to me.

One other thing - do you think we should throw an error if the requested prison is `active: false`? It feels unlikely (and a larger problem) that we'll have any offerings in inactive prisons, but would be a sensible edge case to cover.

server/testutils/factories/organisationAddress.ts
yndajas: Of 128 active prisons, these are the `null` counts:
- 7 on `addressLine1`
- 83 on `addressLine2`
- 14 on `county`

So 5-65% depending on the field

yndajas: But yeah, I do wonder if it's best to keep the default predictable and happy?

server/utils/organisationUtils.test.ts
```diff
@@ -10,116 +12,174 @@ import {
...
+        addressLine2: 'Thine District',
+        town: 'That Town Over There',
+        county: 'Thisshire',
+        postalCode: 'HE3 3TA',
+      }),
+    })
+
+    const email = `nobody-hmp-what@digital.justice.gov.uk`
+
+    describe('when all field are present', () => {
```
Gweaton, line 131: typo on field(s)

Gweaton (approved):
Nice one, thanks. Have tested this out locally and all seems to be working fine.

Looks like you've got a fixup commit here (relevant to our recent conversation!) that'll need squashing in before merging - you may have left it in to make it easier for me to see the changes, if so, thanks!

PR #72, opened 2023-06-16
Use govuk heading classes

yndajas:
## Context

We weren't using the GOV.UK Design System heading classes

## Changes in this PR

- Use GOV.UK Design System heading classes to bring things in line with the standard
- Remove redundant `p` class (inherited from `main`)
- Use class selector where it makes sense

For the heading classes, I've gone with `h1` = `l` and `h2` = `m` since we did have one `h1` that was an `l`. We could go `xl` and `l` instead though. On Approved Premises there are 62 `h1`s with `l` and 7 with `xl`, and 18 `h2`s with `m`, 4 with `l` and 4 with `s` (all `h3`s are `m`)*, so it's contextual (per the GDS guidelines) and/or inconsistent

\* this is based on searching `h1 class="govuk-heading-l"` etc - there may be some heading tags that aren't in that format

#### Before

image

#### After

image

#### Before

image

#### After

image

#### Before

image

#### After

image

## Post-merge checklist

- [ ] Have you written an end-to-end test for the happy path in the [Accredited Programmes E2E repo](https://github.com/ministryofjustice/hmpps-accredited-programmes-e2e) (if necessary)?

### DIFF
--- a/integration_tests/mockApis/auth.ts
+++ b/integration_tests/mockApis/auth.ts
@@ -61,7 +61,7 @@ const redirect = () =>
         'Content-Type': 'text/html',
         Location: 'http://localhost:3007/sign-in/callback?code=codexxxx&state=stateyyyy',
       },
-      body: '<html><body>SignIn page<h1>Sign in</h1></body></html>',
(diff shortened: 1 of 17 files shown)


Gweaton (approved):
Looks good, thanks for this

PR #74, opened 2023-06-20
Fix logger app name

yndajas:


Gweaton (approved):
Thank you! very satisfying

PR #80, opened 2023-06-21
(492) Get docker-compose working locally with new DB

Gweaton:
## Context

https://trello.com/c/VDp60gWQ/492-fix-ui-version-of-the-api-now-that-a-database-has-a-been-added

`script/server` started failing once we added the database to the API, as we weren't starting a postgres database.

## Changes in this PR

This starts all the necessary backing services for running the API and our dependencies locally now that the API has a `seed` environment to start up with data.

## Post-merge checklist

- [ ] This adds or extends a feature...
  - [ ] ... and I have notifed the API developer(s) of changes to the contract tests (Pact), or the API is already compliant

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,6 +23,23 @@ services:
       - SPRING_PROFILES_ACTIVE=dev
       - APPLICATION_AUTHENTICATION_UI_ALLOWLIST=0.0.0.0/0
 
+  postgresql:
+    image: postgres
+    container_name: postgres
+    networks:
+      - hmpps
+    restart: always
+    ports:
+      - '5432:5432'
+    healthcheck:
+      test: [ "CMD", "pg_isready", "--username=admin", "--dbname=postgres"]
+      interval: 10s
+      timeout: 5s
+      retries: 5
+    environment:
+      - POSTGRES_USER=admin
+      - POSTGRES_PASSWORD=admin_password
+
   hmpps-accredited-programmes-api:
     image: quay.io/hmpps/hmpps-accredited-programmes-api:latest
     restart: always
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -62,7 +86,9 @@ services:
     build: .
     networks:
       - hmpps
-    depends_on: [redis]
+    depends_on:
+      - redis
+      - hmpps-accredited-programmes-api
```
yndajas, line 91: Why does this rely on `hmpps-accredited-programmes-api`?

yndajas (approved):
Working great locally for me - just one query

docker-compose.yml
Gweaton: This was used to start the app with docker compose, so I wanted it to wait for the API to start so we could fetch courses, but we never actually use docker compose to start the app. I've removed this whole section now.

script/utils/start-backing-services
```diff
@@ -10,12 +10,12 @@ do
...
-  docker-compose up --scale=app=0 --scale=hmpps-accredited-programmes-api=0 -d
+  docker-compose up --scale=hmpps-accredited-programmes-api=0 -d
 
   echo "==> Stubbing API endpoints..."
 
   npm run api-stubs:reset > /dev/null
   npm run api-stubs:create > /dev/null
 else
-  docker-compose up --scale=app=0 --scale=wiremock=0 -d
+  docker-compose up --scale=wiremock=0 -d
```
yndajas: Not sure I fully understand the commit message here, but basically we were always setting `--scale=app=0` when running `script/server` so it was never being used?

yndajas (approved):
Can we scale postgres down when mocking the API?

Otherwise, looks good

script/utils/start-backing-services
Gweaton: Yeah, exactly. I've reworded the commit message to hopefully be a bit clearer on this, hope it makes more sense.

Gweaton:
> Can we scale postgres down when mocking the API?

Good idea, I've pushed a new commit for this.

PR #81, opened 2023-06-21
(517) Fix prerequisite display order

yndajas:
## Context

We want the prerequisites displayed in a specific order, per the designs. Currently they're displayed in the order in which we get them from the API

## Changes in this PR

- Update prerequisite names per current data
- Fix the order in which they're displayed

### Before

image

### After

image

## Post-merge checklist

- [x] This adds or extends a feature...


yndajas:
[Update courses and offerings pages to match the latest designs](https://trello.com/c/KmLaigqc/517-update-courses-and-offerings-pages-to-match-the-latest-designs)

server/utils/courseUtils.ts
```diff
@@ -21,12 +21,25 @@ const audienceTags = (audiences: Array<CourseAudience>): Array<Tag> => {
 }
 
 const prerequisiteSummaryListRows = (prerequisites: Array<CoursePrerequisite>): Array<SummaryListRow> => {
-  return prerequisites.map(prerequisite => {
-    return {
+  const order: { [key: CoursePrerequisite['name']]: number } = {
```
Gweaton, line 24: I like this approach of using a object with keys rather than relying on the order of an array. I wonder if a `sort` here might be a bit easier to read here, what do you think? Something like:

```
  const order: { [key: CoursePrerequisite['name']]: number } = {
    Setting: 0,
    Gender: 1,
    'Risk criteria': 2,
    'Learning needs': 3,
  }

  const orderedPrerequisites = [...prerequisites].sort(
    (prerequisiteA, prerequisiteB) => order[prerequisiteA.name] - order[prerequisiteB.name],
  )

  return orderedPrerequisites.map(prerequisite => ({
    key: { text: prerequisite.name },
    value: { text: prerequisite.description },
  }))
```

Gweaton (approved):
This looks good to me - one optional suggestion

server/utils/courseUtils.ts
Gweaton: I think your approach might be a bit more robust if a `name` or `description` field is missing from a prerequisite in the API response - I'm not sure how likely that will be if we're managing the data ourselves. Yours strips the incomplete prerequisite from the UI, mine renders the empty `name`/`description` without its counterpart, so happy to take either approach here!

yndajas: Yeah I guess it's a balance between clarity and optimisation. I guess with my approach the purpose of the `forEach` is both sorting and restructuring, but it's less explicit without the `.sort` method. I guess the `index` variable adds some level of explicitness though

A loop would be _O_(_n_<sup>2</sup>) whereas the single index operation is _O_(_n_), although given we have at least one loop with either approach and 2<i>O</i>(_n_<sup>2</sup>) isn't meaningfully different to _O_(_n_<sup>2</sup>), the difference is likely negligible

So no strong feelings - any preference?

Gweaton: Let's go with yours, there's not a lot in it and it's not overly complex either.

PR #84, opened 2023-06-22
(516) Display acronyms

yndajas:
## Context

We want to display an initialised form of course names alongside the course name

## Changes in this PR

This adds `alternateName` to courses and inserts it in the UI

### Before

Note: this does show the initialised forms on the second and third courses, because they were in the `name` in our stubs (removed in this PR), whereas in the screenshot below, they are pulled from `alternateName`

image

### After

image

## Post-merge checklist

### DIFF
--- a/server/@types/models/Course.ts
+++ b/server/@types/models/Course.ts
@@ -4,6 +4,7 @@ import type { CoursePrerequisite } from './CoursePrerequisite'
 export type Course = {
   id: string
   name: string
+  alternateName: string | null
   description: string
   audiences: Array<CourseAudience>
   coursePrerequisites: Array<CoursePrerequisite>
--- a/server/utils/courseUtils.ts
+++ b/server/utils/courseUtils.ts
@@ -43,8 +43,11 @@ const prerequisiteSummaryListRows = (prerequisites: Array<CoursePrerequisite>): 
 }
 
 const presentCourse = (course: Course): CoursePresenter => {
+  const nameAndAlternateName = course.alternateName ? `${course.name} (${course.alternateName})` : course.name
+
   return {
     ...course,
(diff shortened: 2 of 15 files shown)


yndajas:
[Display Acronyms in UI](https://trello.com/c/fwFOGQcJ/516-display-acronyms-in-ui)

server/@types/models/Course.ts
```diff
@@ -4,6 +4,7 @@ import type { CoursePrerequisite } from './CoursePrerequisite'
 export type Course = {
   id: string
   name: string
+  alternateName: string
```
Gweaton, line 7: Sorry, I should have pointed this out on the ticket, but this can be optional, as some courses don't have acronyms e.g. `iHorizon`.

server/utils/courseUtils.ts
```diff
@@ -45,6 +45,7 @@ const prerequisiteSummaryListRows = (prerequisites: Array<CoursePrerequisite>):
 const presentCourse = (course: Course): CoursePresenter => {
   return {
     ...course,
+    nameAndAlternateName: `${course.name} (${course.alternateName})`,
```
Gweaton, line 48: as mentioned above, we'll want to only add the ` (${course.alternateName})` if the course has an acronym

Gweaton (commented):
The approach looks good to me so far.

Apologies, I realise I didn't specify this on the ticket some courses might not have an acronym, so we'll want to add functionality for them not being present. Let's assume they come back as `null` if they're not present.

yndajas:
@Gweaton I've left a squash commit in to show what's changed

Gweaton (approved):
👍 thanks for making that change, glad it wasn't too big of an update!

PR #85, opened 2023-06-23
Jest and `package.json` updates

yndajas:
## Context

We've got some inconsistencies in our testing structure and wording, particularly from inherited code, but also our own project-specific code

### After

- Show seed on Jest tests
- Move Jest and Nodemon config into dedicated files
- Rename methods for consistency
- **Standardise test structure and wording**

## Post-merge checklist


package.json
```diff
@@ -41,6 +41,7 @@
     "npm": "^9"
   },
   "jest": {
+    "showSeed": true,
```
Gweaton: Interesting, do we currently have flaky tests? Would be good to add a card to make those more stable.

Gweaton (approved):
This is great, thank you!

package.json
yndajas: I haven't spotted any, but my thought was that it's better to have the seed when you spot one, than not have it and not be able to reproduce it (until it randomly pops up again)

server/data/hmppsAuthClient.test.ts
```diff
@@ -45,7 +45,7 @@ describe('hmppsAuthClient', () => {
   })
 
   describe('getUserRoles', () => {
-    it('should return data from api', async () => {
+    it('return data from the API', async () => {
```
yndajas, line 48: `returns`

server/data/tokenVerification.ts
```diff
@@ -19,7 +19,7 @@ function getApiClientToken(token: string) {
```
yndajas, line 19: In the commit message: `updated` -> `updates`

PR #86, opened 2023-06-26
[cancelled] Spike paginated courses

yndajas:
## Context

We may or may not want to paginate the list of courses, depending on page length and load times with real data. For now, we don't want to add pagination functionality to the API, so we're handling it on the fronted with query params and splicing of all the courses returned from the API

## Changes in this PR

- Spike on pagination
- This would need a little abstraction/separation of concerns work, but the basic flow and principles are there and working

### Before

[no pagination]

### After

image

image

image

image

## Post-merge checklist

### DIFF
--- a/server/@types/ui/index.d.ts
+++ b/server/@types/ui/index.d.ts
@@ -44,6 +44,12 @@ type OrganisationWithOfferingEmailPresenter = Organisation & {
   summaryListRows: OrganisationWithOfferingEmailSummaryListRows
 }
 
+type PaginatedArray<T> = {
+  page: number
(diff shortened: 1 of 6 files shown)


Gweaton (commented):
This looks like a solid approach and it makes sense to me. I'm almost hoping we do need it, as I think it'll look really neat. Let's leave this as draft for a while longer until we've done "the big data import"

Gweaton:
This is useful context which we can draw on if we need to do it in future, but I'm closing this PR so it's not hanging around in our list of open PRs indefinitely.

PR #87, opened 2023-06-26
(437) Fix CircleCI artifacts and add Cypress HTML report

yndajas:
## Context

Artifacts were not being saved in CircleCI runs of integration tests

## Changes in this PR

- Fix artifact saving in CircleCI integration tests
- Move test config and results into a more consistent structure
- Add HTML report for integration tests

We possibly don't need the XML or JSON reports if we're happy with the HTML, but I've kept/included them for now ([this MoJ repo](https://github.com/ministryofjustice/hmpps-arn-acceptance-tests) does the same)

### Before

image

### After

image

image

image

image

## Post-merge checklist


yndajas:
[Fix Circle artifact and results storage for integration tests](https://trello.com/c/mfjHLS6H/437-fix-circle-artifact-and-results-storage-for-integration-tests)

.circleci/config.yml
```diff
@@ -220,7 +220,7 @@ workflows:
             - deploy_dev
       - hmpps/deploy_env:
           name: deploy_preprod
-          env: "preprod"
+          env: 'preprod'
```
Gweaton, line 229: 👀

package.json
```diff
@@ -12,7 +12,7 @@
     "compile-sass": "sass --quiet-deps --no-source-map --load-path=node_modules/govuk-frontend --load-path=node_modules/@ministryofjustice/frontend --load-path=. assets/scss/application.scss:./assets/stylesheets/application.css assets/scss/application-ie8.scss:./assets/stylesheets/application-ie8.css --style compressed",
     "copy-views": "cp -R server/views dist/server/",
     "int-test-ui": "cypress open",
-    "int-test": "cypress run --config video=false",
+    "int-test": "cypress run",
```
Gweaton, line 15: Should I be able to see the video in the `test_results/cypress` directory here, or is there any more setup I need to do? I can't seem to find those videos, only the xml reports.

Gweaton (approved):
I think it'd be good to get rid of those XML reports too, unless you can think of a reason we might want to keep them?

Approving, providing video works locally

.circleci/config.yml
yndajas: Oops

package.json
yndajas: You might need to show me what you see

Gweaton: as discussed on slack, I was running the wrong script and failing at the unit tests 🤦

jest.config.ts
```diff
@@ -8,7 +8,7 @@ const config: Config = {
     [
       'jest-junit',
       {
-        outputDirectory: 'test_results/unit/xml',
+        outputFile: 'test_results/unit/xml/results-for-ci.xml',
```
Gweaton, line 11: maybe `results-for-ci-report`?

Gweaton (approved):
I'm sure you're on top of this but don't forget to remove those DONOTMERGEs!

PR #89, opened 2023-06-28
(413) Tidy up indices

yndajas:
## Context

We have a lot of inconsistency in our use of index files: some areas of the code use them all the time, some some of the time, and some not at all

## Changes in this PR

Clean up the use of indices, making their use as consistent as possible without breaking anything (there are a couple of exceptions to avoid this)

## Release checklist

[Release process documentation](../doc/how-to/perform-a-release.md)

As part of our continuous deployment strategy we must ensure that this work is
ready to be released once merged.

### Pre-merge

- [ ] There are changes required to the Accredited Programmes API for this change to work...
  - [ ] ... and they been released to production already

### Post-merge

- [ ] This adds, extends or meaningfully modifies a feature...
  - [ ] ... and I have written or updated an end-to-end test for the happy path in the [Accredited Programmes E2E repo](https://github.com/ministryofjustice/hmpps-accredited-programmes-e2e)
- [ ] This makes new expectations of the API...
  - [ ] ... and I have notified the API developer(s) of changes to the contract tests (Pact), or the API is already compliant
- [ ] [Manually approve](../doc/how-to/perform-a-release.md#releasing-to-the-preprod-environment)
  release to preprod
- [ ] [Manually approve](../doc/how-to/perform-a-release.md#releasing-to-the-production-environment)
  release to prod

PR #93, opened 2023-07-03
WIP Simplify

yndajas:
## Context

This builds on #89

## Changes in this PR

- substantially simplify how we're using controllers, services and clients. From Slack:

>I think there are two main changes here. One is directly importing stuff where it's needed instead of creating a long chain of passing stuff around through method parameters. The other is instantiating controllers/services/clients as and when needed, rather than once and using this single instance everywhere. In some contexts the latter was already happening, e.g. clients through their builders
>
> I'm not 100% sure about the performance implications of the latter - whether (re)instantiation on every request is inefficient. I didn't see much difference running the server locally, but I'm one user/session on a fast machine. It should be possible to retain the former change without the latter - I think we were already creating a single instance of controllers and services and exporting them, so it would just be a case of directly importing those rather than passing them through methods

## Release checklist

### Pre-merge

### Post-merge

- [ ] [Manually approve](../doc/how-to/perform-a-release.md#releasing-to-the-preprod-environment)
  release to preprod
- [ ] [Manually approve](../doc/how-to/perform-a-release.md#releasing-to-the-production-environment)
  release to prod

### DIFF
--- a/server/app.ts
+++ b/server/app.ts
@@ -2,7 +2,6 @@ import express from 'express'
 import createError from 'http-errors'
 import path from 'path'
 
-import type { Controllers } from './controllers'
 import errorHandler from './errorHandler'
 import {
   authorisationMiddleware,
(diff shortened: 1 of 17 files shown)


yndajas:
Not something we're pursuing right now. We could look into it in future though

PR #100, opened 2023-07-04
Use accredited-programmes-events slack channel for releases

Gweaton:
## Context

We're currently notifying the dev channel for releases, but a more dedicated events channel for error and deployment notifications feels more appropriate and will keep the dev channel a more focused place for discussion etc.

## Changes in this PR

Changes the channel for notifications to `#accredited-programmes-events`

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,12 +10,12 @@ parameters:
     type: string
     # Normally team specific alert channel e.g. hmpps_tech_alerts, syscon-alerts, dps_sed_alerts
     # This is to avoid a general alert dumping ground that no-one then monitors
-    default: hmpps-accredited-programmes-dev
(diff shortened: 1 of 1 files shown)


yndajas (approved):
Looks good. I think you mentioned that we have the `hmpps` in the current channel to match the app name for some alerts. Is there a reason this needs the same suffix? Could we simplify here if we're creating a more specific channel?

Gweaton:
> Looks good. I think you mentioned that we have the `hmpps` in the current channel to match the app name for some alerts. Is there a reason this needs the same suffix? Could we simplify here if we're creating a more specific channel?

Yeah good point. I thought I'd stick with the current convention but it would be good to try and get rid of it if possible. Have renamed.

PR #111, opened 2023-07-07
(453) Use page headings as titles and other such niceties

yndajas:
## Context

We set page heading and page titles separately, but generally the heading can be reused in the title. There were some other unDRY patterns within the views

## Changes in this PR

- Default to `applicationName` and `pageHeading` if `pageTitle` is not set
- Remove some custom `pageTitle`s
- Use `nameAndAlternateName` in one place it wasn't being used
- DRY up setting of `mainClasses`
- DRY up `h1` creation

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/server/views/partials/layout.njk
+++ b/server/views/partials/layout.njk
@@ -37,6 +37,10 @@
   {% include "./header.njk" %}
 {% endblock header %}
 
+{% if not mainClasses %}
+  {% set mainClasses = "app-container govuk-body" %}
+{% endif %}
+
 {% block beforeContent %}
   {{ govukPhaseBanner({
     tag: {
--- a/server/views/partials/layout.njk
+++ b/server/views/partials/layout.njk
@@ -2,8 +2,14 @@
 
 {% extends "govuk/template.njk" %}
 
+{% if customPageTitleEnd %}
+  {% set pageTitleEnd = " - " + customPageTitleEnd %}
+{% elif pageHeading %}
+  {% set pageTitleEnd = " - " + pageHeading %}
+{% endif %}
+
 {% block pageTitle %}
-  {{pageTitle | default(applicationName)}}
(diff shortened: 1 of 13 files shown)


yndajas:
[Use pageHeadings as pageTitles](https://trello.com/c/5cCfEXAN/453-use-pageheadings-as-pagetitles)

server/views/partials/layout.njk
```diff
@@ -50,6 +50,14 @@
   }) }}
 {% endblock beforeContent %}
 
+{% block content %}
```
Gweaton, line 53: I think I'd like us to err on the side of duplication here, rather than adding this in.

Although it's essentially the same line of code on each page, I find it easier to consider the HTML structure of the page here (given this is used in non-partial templates) knowing that there's an `<h1>` explicitly declared in this file. 

Having this new block introduces a bit more complexity as to know there's an `<h1>` here you'd have to know to check the `layout.njk` file.

I know this is inconsistent with how we set `title`s but I consider them less of a "main character" in the structure of an HTML page, if that makes any sense at all.

Gweaton (approved):
This looks good to me, but let's remove the last commit for now, unless you have any strong objections?

server/views/partials/layout.njk
yndajas: Happy to undo that. What do you think about the alternative I suggested in the commit message? I think I asked about this in an early PR, but page headings seem to be either an outlier as the hardcoded text content that gets passed down from the controller, or using data that's already available from other stuff that gets passed down (e.g. course or organisation name)

(I'd be more keen on this than the abstraction to setting the heading in the layout actually!)

> Alternatively, we could skip passing the heading down as a variable and move the currently controller-based code into the view (`coursePresenter.nameAndAlternateName`, `"List of accredited programmes"` and `"Accredited Programmes"`)

PR #140, opened 2023-08-07
(642a) Add navigation

yndajas:
## Context

The designs have a navigation menu with "List of programmes" and "My cases". The latter doesn't currently have a corresponding page in our service, but the former does

This is part one of (probably) two PRs to introduce the refer start page

## Changes in this PR

- Add the navigation with a "List of programmes" link
- A few iterations in the same area of code

### Before

Courses page:

image

Course page:

image

Dashboard page (no change):

image

### After

Courses page:

image

Course page:

image

Dashboard page (no change):

image

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/integration_tests/pages/page.ts
+++ b/integration_tests/pages/page.ts
@@ -25,18 +25,26 @@ export default abstract class Page {
     }
   }
 
-  checkOnPage(): void {
+  private checkOnPage(): void {
     if (!this.external) {
-      let expectedTitle = 'HMPPS Accredited Programmes'
-      const pageTitleEnd = this.customPageTitleEnd || this.pageHeading
-
-      if (pageTitleEnd) {
-        expectedTitle += ` - ${pageTitleEnd}`
-      }
-
-      cy.title().should('equal', expectedTitle)
+      this.checkTitle()
     }
 
+    this.checkHeading()
+  }
+
+  private checkTitle(): void {
+    let expectedTitle = 'HMPPS Accredited Programmes'
+    const pageTitleEnd = this.customPageTitleEnd || this.pageHeading
+
+    if (pageTitleEnd) {
+      expectedTitle += ` - ${pageTitleEnd}`
+    }
+
+    cy.title().should('equal', expectedTitle)
+  }
+
+  private checkHeading(): void {
     cy.get('.govuk-heading-l').contains(this.pageHeading)
   }
 
--- a/integration_tests/pages/page.ts
+++ b/integration_tests/pages/page.ts
@@ -83,4 +83,29 @@ export default abstract class Page {
   shouldContainBackLink(href: string): void {
     cy.get('.govuk-back-link').should('have.attr', 'href', href)
   }
+
+  shouldContainNavigation(currentPath: string): void {
+    const navigationItems = [{ text: 'List of programmes', href: '/programmes' }]
+
+    cy.get('.moj-primary-navigation__item').each((navigationItemElement, navigationItemElementIndex) => {
+      const { text, href } = navigationItems[navigationItemElementIndex]
+
+      const { actual, expected } = helpers.parseHtml(navigationItemElement, text)
+      expect(actual).to.equal(expected)
+
+      cy.wrap(navigationItemElement).within(() => {
+        cy.get('a').should('have.attr', 'href', href)
+
+        if (currentPath === href) {
+          cy.get('a').should('have.attr', 'aria-current', 'page')
+        } else {
+          cy.get('a').should('not.have.attr', 'aria-current', 'page')
+        }
+      })
+    })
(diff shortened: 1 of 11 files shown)


yndajas:
[Make a referral start page (UI)](https://trello.com/c/DeqPNrKc/642-make-a-referral-start-page-ui)

server/views/partials/layout.njk
```diff
@@ -53,6 +54,16 @@
 
 {% block header %}
   {% include "./header.njk" %}
+
+  {{ mojPrimaryNavigation({
```
Gweaton: Although you've said in MOJ projects this goes above the phase banner, I think we should follow the GOV.UK design system's "must" [here](https://design-system.service.gov.uk/components/phase-banner/#how-it-works) and put it below:

> Your banner must be directly under the black GOV.UK header and colour bar.

server/views/partials/layout.njk
```diff
@@ -32,7 +32,201 @@
...
+      !function (v, y, T) {
```
Gweaton: Let's either put this in a formatting commit, or undo this change and have it all compressed (personal preference for the latter to reduce the amount of scrolling as this is a script copied from Azure and we probably won't need to debug/tweak it (famous last words...)

integration_tests/pages/page.ts
```diff
@@ -83,4 +83,30 @@ export default abstract class Page {
   shouldContainBackLink(href: string): void {
     cy.get('.govuk-back-link').should('have.attr', 'href', href)
   }
+
+  checkNavigation(currentPath: string, navigationEnabled: boolean): void {
```
Gweaton, line 95: Could we call this something like `shouldShow/ContainNavigation` to make the boolean arg line up a bit more clearly with the function name?

server/views/partials/applicationInsightsScript.njk
```diff
@@ -0,0 +1,209 @@
+<script nonce="{{ cspNonce }}" type="text/javascript">
```
Gweaton, line 1: Ah this addresses my comment from an earlier commit - looks like a good approach to me.

server/views/partials/layout.njk
Gweaton: The issue above is resolved, but could you take this formatting out of this commit please?

Gweaton (commented):
Nice work, looks good to me, just a couple of small tweaks

server/views/partials/layout.njk
yndajas: Isn't that where it is at the moment? I noticed yesterday we're using a HMPPS header variant that doesn't have the colour bar, but at the moment the navigation is directly below the black header

Gweaton: Sorry if it wasn't clear, I was referring to the phase banner above, so black header -> phase banner -> navigation

integration_tests/pages/page.ts
yndajas: My thought was that it would be weird to have a parameter asking whether to show navigation in a method called `shouldShowNavigation`, but on reflection I suppose it does make more sense in the method signature when calling it. Does `navigationEnabled` make sense as the parameter though?

server/views/partials/applicationInsightsScript.njk
yndajas: Actually I wanted to check with you if there's a way we can confirm this is working? Do we already have tests covering that?

server/views/partials/layout.njk
yndajas: Ah yes. I'm not sure if we can do this without hacking the CSS or replacing `block content` with `block postNavigationContent` in each view though, which is similar to something you weren't keen on before 🤔

integration_tests/pages/page.ts
Gweaton: How would you feel about having two functions here, one doing the `!navigationEnabled` section and one doing the other - i.e. `should` and `shouldNot` show navigation? I don't know if there's a particular benefit to having it as one function when the conditionals are kind of doing different things.

server/views/partials/layout.njk
Gweaton: Could we add a new block to the `beforeContent` block after the `govukPhaseBanner` and override it on pages that we don't want it on (which may just be the initial dashboard page at the moment)?

Gweaton: like we do for the `backLink`

yndajas: Yep - it just needs some CSS hacking so that the navigation is full width but I'm working on reusing the CSS from the dashboard

Gweaton (approved):
Thanks for tinkering with that CSS. Have tested out locally and it's looking good!

PR #142, opened 2023-08-10
(642b) Add refer start

yndajas:
## Context

Adding feature flag integration tests will take a lot more work, so we'll do this in a separate PR - https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/commit/1ca9ec408b2ec0bc5c8d93e6d95b4c15d953b736 on the branch `feature-flag-tests` is a spike for this

## Changes in this PR

Add the referral start page and related work

### Before

image

### After

image

image

## Release checklist

### Pre-merge

### Post-merge


yndajas:
[Make a referral start page (UI)](https://trello.com/c/DeqPNrKc/642-make-a-referral-start-page-ui)

server/controllers/refer/referralsController.ts
```diff
@@ -0,0 +1,41 @@
...
+    private readonly courseService: CourseService,
+    private readonly organisationService: OrganisationService,
+  ) {}
+
+  start(): TypedRequestHandler<Request, Response> {
+    return async (req: Request, res: Response) => {
+      typeUtils.assertHasUser(req)
+
+      const course = await this.courseService.getCourse(req.user.token, req.params.courseId)
+      const courseOffering = await this.courseService.getOffering(
```
Gweaton, line 18: Do we need to account for the possibility here of an offering not being found, like we do for Organisations?

server/paths/find.ts
```diff
@@ -3,7 +3,7 @@ import { path } from 'static-path'
 const coursesPath = path('/programmes')
 const coursePath = coursesPath.path(':courseId')
 
-const courseOfferingPath = coursePath.path('offerings/:courseOfferingId')
+export const courseOfferingPath = coursePath.path('offerings/:courseOfferingId')
```
Gweaton, line 6: Can we not use the exported version in the `default` object below?

server/controllers/refer/referralsController.ts
```diff
@@ -0,0 +1,41 @@
...
+        req.user.token,
+        req.params.courseId,
+        req.params.courseOfferingId,
+      )
+      const organisation = await this.organisationService.getOrganisation(req.user.token, courseOffering.organisationId)
+
+      if (!organisation) {
```
Gweaton, line 25: I don't think we've got these yet, but it'd be good to add tests for 404ing in controllers.

server/paths/refer.ts
```diff
@@ -0,0 +1,9 @@
+import { courseOfferingPath } from './find'
+
+const referralStartPath = courseOfferingPath.path('refer')
+
+export default {
+  referrals: {
```
Gweaton, line 6: How would you feel about removing the top level of this, so we'd just have a `start` key. Are there any non-referral based paths we might have in here? Or were you thinking of matching the route structure here?

Gweaton (approved):
This looks good to me - a couple of small things but nothing major.

server/paths/refer.ts
yndajas: I was more matching what we did on find - could we not remove the `courses` key over there by the same principle? Perhaps we should

server/controllers/refer/referralsController.ts
yndajas: We don't in the course offerings controller's show action (or either of the course controller's actions or the offering one for courses). I believe our thinking was that organisations are external so we can't guarantee what comes back from the prison register API, whereas we have more confidence in our own data/API. Of course someone could play with the IDs in the URL but I think we previously decided only to deal with the clients that touch external data. Not sure what happens if we try to access a page using a non-existent offering or course ID

server/controllers/refer/referralsController.ts
yndajas: Can't remember if I had trouble with this when trying the flag stuff but I'll give this a go tomorrow

server/paths/refer.ts
yndajas: What structure did you use for the people show path?

server/controllers/refer/referralsController.ts
Gweaton: I've added one [here](https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/pull/143/commits/9d525a6feaedd8dc70984b0608bc3481f8e28f78) if it's useful

server/paths/refer.ts
Gweaton: Yeah I think it might make sense

server/controllers/refer/referralsController.ts
yndajas: I've added a commit for this: https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/pull/142/commits/7bfb50a6e9fa0582da2fa9bb88999730e3b07dfa

I found that adding the object after `404` in `createError` made no difference to whether the test passed - you could put any object in there and it would be happy - so I simplified it a bit from your example

Gweaton: Ah interesting, good to know - thanks for updating this.

Gweaton (approved):

PR #155, opened 2023-08-22
[680] Add user details to Application Insights requests

Gweaton:
## Context

https://trello.com/c/5AhnbNSG/680-app-insights-changes

## Changes in this PR

Adds username and active case load ID to the request object we send to Application Insights. This allows for more reliable user metrics than the auto-generated user ID we currently see in Application Insights which is attributing multiple users to the same user account when logging in e.g. in another browser/session.

This is heavily based off [work done by the Use of Force team](https://github.com/ministryofjustice/use-of-force/pull/599/files) with a few more manual typings required, as well as me adding a bit more description to test cases.

This also converts the AppInsights utils file into a class so we can make use of our function sorting.

## Release checklist

### Pre-merge

### Post-merge


server/utils/appInsightsUtils.ts
```diff
@@ -1,13 +1,43 @@
...
+}
+
+export type ContextObject = Record<string, ContextObjectWithUser> | undefined
 
 export default class AppInsightsUtils {
+  static addUserDataToRequests = (envelope: EnvelopeTelemetry, contextObjects: ContextObject): boolean => {
+    const isRequest = envelope.data.baseType === Contracts.TelemetryTypeString.Request
+    if (isRequest) {
+      const { username, activeCaseLoadId } = contextObjects?.['http.ServerRequest']?.res?.locals?.user || {}
+      if (username && envelope.data.baseData) {
```
yndajas, line 23: Could we condense this into one conditional? The only benefit of having two is not reading two properties, and it would flatten this block a bit/make it more readable

server/utils/appInsightsUtils.ts
```diff
@@ -1,13 +1,43 @@
...
+        const { properties } = envelope.data.baseData
+        // eslint-disable-next-line no-param-reassign
+        envelope.data.baseData.properties = {
+          activeCaseLoadId,
+          username,
+          ...properties,
+        }
```
yndajas, line 30: I'm guessing `activeCaseLoadId` and `username` wouldn't have been in the original `envelope.data.baseData.properties`? (So they won't be overwritten by `...properties` coming at the end rather than the start?

server/utils/appInsightsUtils.ts
```diff
@@ -1,13 +1,43 @@
...
+      }
+    }
+    return true
+  }
+
   static buildClient = (name = this.defaultName()): TelemetryClient | null => {
     if (process.env.APPLICATIONINSIGHTS_CONNECTION_STRING) {
       defaultClient.context.tags['ai.cloud.role'] = name
       defaultClient.context.tags['ai.application.ver'] = this.version()
+      defaultClient.addTelemetryProcessor(this.addUserDataToRequests)
```
yndajas, line 40: Does `this.` work here? I tried that before in another newly classed file and ended up using `ClassName.` instead since `this.`, which wasn't working as I'd expected from one class method to another

server/utils/appInsightsUtils.ts
```diff
@@ -1,13 +1,43 @@
...
+
+type ContextObjectWithUser = {
+  res?: {
+    locals?: {
+      user: Partial<User>
+    }
+  }
+}
+
+export type ContextObject = Record<string, ContextObjectWithUser> | undefined
```
yndajas, line 16: A `ContextObject` is an object with key/value pairs with strings pointing to `ContextObjectWithUser`s? So a context object includes context objects as its keys' values?

Below your argument of `contextObjects` (plural) is a `ContextObject` - perhaps this type should be `ContextObjects`? Are there going to be multiple `ContextObjectWithUser`s in any `ContextObject`s?

server/data/index.ts
```diff
@@ -3,10 +3,10 @@
  * Do appinsights first as it does some magic instrumentation work, i.e. it affects other 'require's
  * In particular, applicationinsights automatically collects bunyan logs
  */
-import { buildAppInsightsClient, initialiseAppInsights } from '../utils/azureAppInsights'
+import AppInsightsUtils from '../utils/appInsightsUtils'
```
yndajas, line 6: Is it worth seeing if you can import/export from the utils index now that it's a class? It didn't work before, but this could have fixed it?

server/utils/appInsightsUtils.test.ts
```diff
@@ -0,0 +1,86 @@
...
+
+const user = {
+  activeCaseLoadId: 'MDI',
+  username: 'TEST_USER',
+}
+
+const createEnvelope = (properties: Record<string, string | boolean> | undefined, baseType = 'RequestData') =>
+  ({
+    data: {
+      baseData: { properties },
```
yndajas, line 14: Wouldn't properties already be an object?

yndajas (commented):
A few questions - it may be that I'm not understanding a few things here

server/data/hmppsAuthClient.ts
```diff
@@ -34,6 +34,7 @@ function getSystemClientTokenFromHmppsAuth(username?: string): Promise<superagen
 interface User {
   activeCaseLoadId: string
   name: string
+  username: string
```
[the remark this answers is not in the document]
Gweaton: The auto-generated user ID we currently see in Application Insights is attributing multiple users to the same user account when logging in e.g. in another browser/session.

server/utils/appInsightsUtils.test.ts
```diff
@@ -0,0 +1,86 @@
...
+            activeCaseLoadId,
+            username,
+          },
+        },
+      },
+    },
+  }) as ContextObject
+
+describe('AppInsightsUtils', () => {
+  describe('addUserDataToRequests', () => {
```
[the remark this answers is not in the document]
Gweaton: Yeah, I don't think we can test much of it without some rather complex mocking of the black box AppInsights module. I've not seen other teams manage to test it

server/data/hmppsAuthClient.ts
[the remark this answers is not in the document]
yndajas: I think it would be good - I'd assume a change in this area would be to do with anonymity, but switching user ID for username doesn't do that

server/utils/appInsightsUtils.test.ts
```diff
@@ -0,0 +1,86 @@
...
+      const envelope = createEnvelope({ other: 'things' })
+
+      AppInsightsUtils.addUserDataToRequests(envelope, {
+        'http.ServerRequest': {},
+      } as ContextObject)
+
+      expect(envelope.data.baseData!.properties).toEqual({
+        other: 'things',
+      })
+    })
```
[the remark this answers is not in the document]
Gweaton: Essentially all of the optional chaining with `contextObjects?.['http.ServerRequest']?.res?.locals?.user`

server/data/index.ts
Gweaton: As discussed, will leave this for now.

yndajas (approved):

PR #156, opened 2023-08-22
(704) Create referral

yndajas:
## Context

We have pages for confirming a person and showing the referral tasklist. In between we need to create the referral and redirect from the former to the latter

## Changes in this PR

- Create a referral on confirming person details, then redirect to the tasklist

## Release checklist

### Pre-merge

- [x] There are changes required to the Accredited Programmes API for this change to work...
  - [x] ... and they have been released to production already

### Post-merge

  - [x] ... and I have notified the API developer(s) of changes to the contract tests (Pact), or the API is already compliant

### DIFF
--- a/server/@types/models/Referral.ts
+++ b/server/@types/models/Referral.ts
@@ -1,6 +1,12 @@
-export type Referral = {
+type Referral = {
   id: string // eslint-disable-next-line @typescript-eslint/member-ordering
   offeringId: string
   prisonNumber: string
   referrerId: string
 }
+
+type CreatedReferralResponse = {
+  referralId: Referral['id']
+}
+
+export type { CreatedReferralResponse, Referral }
--- a/server/@types/models/Referral.ts
+++ b/server/@types/models/Referral.ts
@@ -1,8 +1,11 @@
+import type { CourseOffering } from './CourseOffering'
+import type { Person } from './Person'
+
 type Referral = {
   id: string // eslint-disable-next-line @typescript-eslint/member-ordering
-  offeringId: string
-  prisonNumber: string
-  referrerId: string
+  offeringId: CourseOffering['id']
+  prisonNumber: Person['prisonNumber']
+  referrerId: Express.User['userId']
 }
 
 type CreatedReferralResponse = {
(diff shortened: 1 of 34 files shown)


server/@types/models/Referral.ts
```diff
@@ -1,6 +1,12 @@
-export type Referral = {
+type Referral = {
   id: string // eslint-disable-next-line @typescript-eslint/member-ordering
   offeringId: string
   prisonNumber: string
   referrerId: string
 }
+
+type ReferralIdWrapper = {
```
Gweaton, line 11: Maybe `CreatedReferralResponse`? Might make it a bit more obvious what its use case is

server/@types/models/Referral.ts
```diff
@@ -1,8 +1,11 @@
...
+import type { Person } from './Person'
+
 type Referral = {
   id: string // eslint-disable-next-line @typescript-eslint/member-ordering
-  offeringId: string
-  prisonNumber: string
-  referrerId: string
+  offeringId: CourseOffering['id']
+  prisonNumber: Person['prisonNumber']
+  referrerId: Express.User['userId'] // no idea which of the three User/UserDetails interfaces is best to use here
```
Gweaton, line 8: I think this one is probably fine - I was going to suggest using the `User` model but that doesn't have the `token` on it yet. This one'll do us

server/paths/api.ts
```diff
@@ -18,4 +20,7 @@ export default {
     },
     show: coursePath,
   },
+  referrals: {
+    index: referralsPath,
```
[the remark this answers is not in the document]
yndajas: I suppose. It would be the same path if we had an index though. Would we duplicate the path with an `index` key if we had one? The path/route distinction in Express is a little different to Rails 🤔

server/@types/models/Referral.ts
yndajas: Yeah, that's a bit less nonchalant!

server/paths/api.ts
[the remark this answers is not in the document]
Gweaton: Yeah I think we probably would duplicate them to distinguish from posting to create and getting multiple results back

PR #148, opened 2023-08-15
(643) Find a person

yndajas:
## Context

We have the referral start page and the person show page, but we're missing the referral new/person search page in the middle

## Changes in this PR

- Add the referral new/person search page
- Various related iterations

## Screenshots of UI changes

https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/assets/40244233/633cada1-70f1-4ffa-98b0-676b2fa8a22a

## Release checklist

[Release process documentation](../doc/how-to/perform-a-release.md)

As part of our continuous deployment strategy we must ensure that this work is
ready to be released once merged.

### Pre-merge

- [ ] There are changes required to the Accredited Programmes API for this change to work...
  - [ ] ... and they have been released to production already

### Post-merge

- [x] This adds, extends or meaningfully modifies a feature...
  - [ ] ... and I have written or updated an end-to-end test for the happy path in the [Accredited Programmes E2E repo](https://github.com/ministryofjustice/hmpps-accredited-programmes-e2e)
- [ ] This makes new expectations of the API...
  - [ ] ... and I have notified the API developer(s) of changes to the contract tests (Pact), or the API is already compliant
- [ ] [Manually approve](../doc/how-to/perform-a-release.md#releasing-to-the-preprod-environment) release to preprod
- [ ] [Manually approve](../doc/how-to/perform-a-release.md#releasing-to-the-production-environment) release to prod

PR #158, opened 2023-08-23
(732) Test joined up journey for looking up a person in prison from the Prisoner Offender Search

Gweaton:
## Context

https://trello.com/c/TovAzIoK/732-add-integration-test-for-searching-for-a-user-when-making-a-referral

We recently #148 added functionality for finding a person to be referred to an Accredited programme. The journey for finding a user and then showing them were added in piecemeal PRs, so we neglected to test the whole journey as a joined-up user experience.

## Changes in this PR

Combines integration tests to give us a more complete user journey when testing the application.

We're only testing the happy path here, as we've not yet got any custom behaviour for showing error messages if a user isn't found, but we'll add test cases to cover that scenario when we add those error messages in the near future.

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/integration_tests/e2e/refer.cy.ts
+++ b/integration_tests/e2e/refer.cy.ts
@@ -42,26 +42,7 @@ context('Refer', () => {
     startReferralPage.shouldContainStartButtonLink()
   })
 
-  it("Shows the 'find person' page for a referral", () => {
-    cy.signIn()
-
-    const course = courseFactory.build()
-    const courseOffering = courseOfferingFactory.build()
-
-    cy.task('stubCourseByOffering', { course, courseOfferingId: courseOffering.id })
-    cy.task('stubCourseOffering', { courseId: course.id, courseOffering })
-
-    const path = referPaths.new({ courseOfferingId: courseOffering.id })
-    cy.visit(path)
-
-    const findPersonPage = Page.verifyOnPage(FindPersonPage)
-    findPersonPage.shouldContainNavigation(path)
-    findPersonPage.shouldContainBackLink(referPaths.start({ courseOfferingId: courseOffering.id }))
-    findPersonPage.shouldContainInstructionsParagraph()
-    findPersonPage.shouldContainIdentifierForm()
-  })
-
-  it("Shows the 'confirm person' page when starting a new referral", () => {
+  it("Allows users to search for a person and confirm that person's details", () => {
     cy.signIn()
 
     const course = courseFactory.build()
@@ -71,6 +52,21 @@ context('Refer', () => {
       firstName: 'Del',
       lastName: 'Hatton',
     })
+
+    cy.task('stubCourseByOffering', { course, courseOfferingId: courseOffering.id })
+    cy.task('stubCourseOffering', { courseId: course.id, courseOffering })
+    cy.task('stubPrisoner', prisoner)
+
+    const path = referPaths.new({ courseOfferingId: courseOffering.id })
+    cy.visit(path)
+
+    const findPersonPage = Page.verifyOnPage(FindPersonPage)
+    findPersonPage.shouldContainNavigation(path)
+    findPersonPage.shouldContainBackLink(referPaths.start({ courseOfferingId: courseOffering.id }))
+    findPersonPage.shouldContainInstructionsParagraph()
+    findPersonPage.shouldContainIdentifierForm()
+
+    findPersonPage.searchForPerson(prisoner.prisonerNumber)
 
     const person = personFactory.build({
       currentPrison: prisoner.prisonName,
@@ -83,13 +79,6 @@ context('Refer', () => {
       setting: 'Custody',
     })
 
-    cy.task('stubPrisoner', prisoner)
-
-    const path = referPaths.people.show({
-      courseOfferingId: courseOffering.id,
-      prisonNumber: person.prisonNumber,
-    })
-    cy.visit(path)
     const confirmPersonPage = Page.verifyOnPage(ConfirmPersonPage, { course, courseOffering, person })
 
     confirmPersonPage.shouldContainNavigation(path)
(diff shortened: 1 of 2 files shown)


integration_tests/pages/refer/findPerson.ts
```diff
@@ -1,11 +1,18 @@
 import Helpers from '../../support/helpers'
 import Page from '../page'
+import type { Prisoner } from '@prisoner-offender-search'
 
 export default class FindPersonPage extends Page {
   constructor() {
     super("Enter the person's identifier")
   }
 
+  searchForPrisonerNumber(prisonerNumber: Prisoner['prisonerNumber']) {
```
yndajas, line 10: Would it be better to use our internal language rather than the POS language: `searchForPerson(prisonNumber: Person['prisonNumber'])`?

integration_tests/pages/refer/findPerson.ts
```diff
@@ -1,11 +1,18 @@
...
+    cy.get('input#prisonNumber').type(prisonerNumber)
+
+    cy.contains('.govuk-button', 'Continue').click()
```
yndajas, line 13: You could use the `shouldContainButton` method you wrote here? `this.shouldContainButton('Continue').click()`

integration_tests/e2e/refer.cy.ts
```diff
@@ -83,13 +79,6 @@ context('Refer', () => {
       setting: 'Custody',
     })
```
yndajas, line 80: I can't remember where we got to in our thinking around using utils in tests (was that about unit tests?) but here we could just say `const person = PersonUtils.personFromPrisoner(prisoner)` rather than recreating it?

It was like this before, so maybe we concluded this was better?

integration_tests/e2e/refer.cy.ts
```diff
@@ -42,14 +42,20 @@ context('Refer', () => {
     startReferralPage.shouldContainStartButtonLink()
   })
 
-  it("Shows the 'find person' page for a referral", () => {
+  it("Allows users to search for a person and confirm that person's details", () => {
```
yndajas, line 45: So are we favouring testing two pages together where they combine to form a discreet part of the overall journey? I guess this makes it easier to test the search action without duplicating most of the find person page logic in a discreet test

I suppose though I'll end up needing to duplicate some of the confirm person page logic from this test when testing the create and referral show actions in my PR?

yndajas (approved):
A few suggestions but nothing substantial. Would be good to chat about the final comment re: implications for my PR

integration_tests/e2e/refer.cy.ts
yndajas: (Or, more concisely: we will have a lot of on-click/between page functionality - so I wonder how we cut up our integration tests)

integration_tests/pages/refer/findPerson.ts
```diff
@@ -1,11 +1,18 @@
...
 import Page from '../page'
+import type { Prisoner } from '@prisoner-offender-search'
 
 export default class FindPersonPage extends Page {
   constructor() {
     super("Enter the person's identifier")
   }
 
+  searchForPrisonerNumber(prisonerNumber: Prisoner['prisonerNumber']) {
+    cy.get('input#prisonNumber').type(prisonerNumber)
```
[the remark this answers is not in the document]
Gweaton: Yeah I thought this was a bit clearer to read

integration_tests/pages/refer/findPerson.ts
Gweaton: I went back and forth on this, as it's the Prisoner Offender Search we're searching in (this is the trouble with using two very similar but different terms!) but happy to update

integration_tests/e2e/refer.cy.ts
Gweaton: Yeah I think I wanted to test the whole integration here without relying on our internal implementation, hence creating the expected Person

integration_tests/pages/refer/findPerson.ts
yndajas: Well, we're testing that our service is able to search for something, no? Should the integration tests know how it does it? Are the integration tests not abstracted away from the 'how'?

integration_tests/e2e/refer.cy.ts
yndajas: Makes sense - I guess this strengthens the case for avoiding POS language above?

integration_tests/e2e/refer.cy.ts
Gweaton: Yeah, integration test-wise, I think it's useful to chop things up into smaller discreet parts as you've mentioned, rather than testing a whole user journey. In our E2E tests we can test the joined up happy path for the whole journey.

For the create/show part of your PR, I think we could visit the confirm person page for a person, not bother about asserting anything, then click "Continue" and assert that see the created Referral.

yndajas: Yeah, I think that's where I was heading. Thanks!

integration_tests/e2e/refer.cy.ts
Gweaton: Yep, have switched that above

integration_tests/pages/refer/findPerson.ts
Gweaton: Have updated this to use the term `person`. There's a slight strangeness that we're passing in a `prisoner.prisonerNumber` when we define the type of the parameter as `Person['prisonNumber']` but under the hood they're the same thing so I don't think it's a problem. It also hides that implementation detail from users, as you say

yndajas: Could the person not be created further up in the test, then you can pass the person's prison number instead?

Gweaton: I've merged this as-is as the "Person" doesn't exist at that point in our service and users will be reading the prison number from a record in DPS/NOMIS to enter it here and I found this to be a bit more of a logical flow.

PR #170, opened 2023-09-04
(651) Check answers page for referrals

jsrobertson:
## Changes in this PR
New check answers page for referrals.  Contains application summary and personal details summary lists.

### Before
N/A

### After
localhost_3000_referrals_c1ae9443-0a50-467a-9413-07c88549e1fa_check-answers

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/integration_tests/e2eReferDisabled/refer.cy.ts
+++ b/integration_tests/e2eReferDisabled/refer.cy.ts
@@ -86,4 +86,26 @@ context('Refer', () => {
     const notFoundPage = Page.verifyOnPage(NotFoundPage)
     notFoundPage.shouldContain404H2()
   })
+
+  it("Doesn't show the check answers page for a referral", () => {
+    cy.signIn()
+
+    const course = courseFactory.build()
+    const courseOffering = courseOfferingFactory.build()
+    const prison = prisonFactory.build({ prisonId: courseOffering.organisationId })
+    const prisoner = prisonerFactory.build()
+    const referral = referralFactory.build({ offeringId: courseOffering.id, prisonNumber: prisoner.prisonNumber })
+
+    cy.task('stubCourseByOffering', { course, courseOfferingId: courseOffering.id })
+    cy.task('stubCourseOffering', { courseId: course.id, courseOffering })
+    cy.task('stubPrison', prison)
+    cy.task('stubPrisoner', prisoner)
+    cy.task('stubReferral', referral)
+
+    const path = referPaths.checkAnswers({ referralId: referral.id })
+    cy.visit(path, { failOnStatusCode: false })
+
+    const notFoundPage = Page.verifyOnPage(NotFoundPage)
+    notFoundPage.shouldContain404H2()
+  })
 })
--- a/integration_tests/mockApis/auth.ts
+++ b/integration_tests/mockApis/auth.ts
@@ -4,6 +4,8 @@ import type { Response } from 'superagent'
 import tokenVerification from './tokenVerification'
 import { getMatchingRequests, stubFor } from '../../wiremock'
 
+const mockedUsername = 'USER1'
+
 const createToken = () => {
   const payload = {
     auth_source: 'nomis',
@@ -11,7 +13,7 @@ const createToken = () => {
     client_id: 'clientid',
     jti: '83b50a10-cca6-41db-985f-e87efb303ddb',
     scope: ['read'],
-    user_name: 'USER1',
+    user_name: mockedUsername,
   }
 
   return jwt.sign(payload, 'secret', { expiresIn: '1h' })
@@ -112,7 +114,7 @@ const token = () =>
         internalUser: true,
         scope: 'read',
         token_type: 'bearer',
-        user_name: 'USER1',
+        user_name: mockedUsername,
       },
       status: 200,
     },
@@ -132,7 +134,7 @@ const stubUser = (name: string) =>
         active: true,
         name,
         staffId: 231232,
-        username: 'USER1',
+        username: mockedUsername,
       },
       status: 200,
     },
@@ -155,6 +157,7 @@ const stubUserRoles = () =>
 
 export default {
   getSignInUrl,
+  mockedUsername,
   stubAuthPing: ping,
   stubAuthUser: (name = 'john smith'): Promise<[Response, Response]> => Promise.all([stubUser(name), stubUserRoles()]),
   stubSignIn: (): Promise<[Response, Response, Response, Response, Response, Response]> =>
--- a/server/controllers/refer/referralsController.test.ts
+++ b/server/controllers/refer/referralsController.test.ts
@@ -7,13 +7,14 @@ import ReferralsController from './referralsController'
 import { referPaths } from '../../paths'
 import type { CourseService, OrganisationService, PersonService, ReferralService } from '../../services'
 import {
+  courseAudienceFactory,
   courseFactory,
   courseOfferingFactory,
   organisationFactory,
   personFactory,
   referralFactory,
 } from '../../testutils/factories'
-import { CourseUtils, ReferralUtils, TypeUtils } from '../../utils'
+import { CourseUtils, PersonUtils, ReferralUtils, TypeUtils } from '../../utils'
 import type { CoursePresenter } from '@accredited-programmes/ui'
 
 jest.mock('../../utils/courseUtils')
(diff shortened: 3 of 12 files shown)


server/controllers/refer/referralsController.test.ts
```diff
@@ -145,4 +146,42 @@ describe('ReferralsController', () => {
...
+
+      const referral = referralFactory.build({ offeringId: courseOffering.id, prisonNumber: person.prisonNumber })
+      referralService.getReferral.mockResolvedValue(referral)
+
+      request.params.referralId = referral.id
+
+      const coursePresenter = createMock<CoursePresenter>({
+        audiences: courseAudienceFactory.buildList(1),
+        nameAndAlternateName: `${course.name} (${course.alternateName})`,
+      })
```
yndajas, line 166: I know we're doing this elsewhere, but I wonder what the use is of setting these values on the mocked `coursePresenter` when we aren't asserting on them. Probably another conversation for later!

server/controllers/refer/referralsController.test.ts
```diff
@@ -145,4 +146,42 @@ describe('ReferralsController', () => {
...
+        applicationSummaryListRows: ReferralUtils.applicationSummaryListRows(
+          courseOffering,
+          coursePresenter,
+          organisation,
+          person,
+        ),
+        courseOfferingId: courseOffering.id,
+        pageHeading: 'Check your answers',
+        person,
+        personSummaryListRows: PersonUtils.summaryListRows(person),
```
yndajas, line 182: We might be at the point that it would be useful to have a `PersonPresenter` rather than a public `personSummaryListRows` method (since we need to send both the person and their summary list here), like with `CoursePresenter` and `OrganisationWithOfferingEmailsPresenter`, but again one for another time

server/views/referrals/checkAnswers.njk
```diff
@@ -0,0 +1,103 @@
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+{% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
+{% from "govuk/components/checkboxes/macro.njk" import govukCheckboxes %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+{% from "./_taskListSection.njk" import taskListSection %}
+{% from "../partials/audienceTags.njk" import audienceTags %}
```
yndajas, line 7: Don't need either of these here

server/views/referrals/checkAnswers.njk
```diff
@@ -0,0 +1,103 @@
...
+
+            <hr class="govuk-section-break govuk-section-break--visible govuk-!-margin-bottom-6">
+
+            <h2 class="govuk-heading-m">Application summary</h2>
+
+            {{ govukSummaryList({
+                classes: 'govuk-summary-list--no-border',
+                rows: applicationSummaryListRows,
+                attributes: {
+                    "data-testid": "application-summary-list"
```
yndajas, line 37: Nice! I've added a card for us to do this for other stuff we target in integration tests

server/views/referrals/checkAnswers.njk
```diff
@@ -0,0 +1,103 @@
...
+                    {{ govukButton({
+                        text: "Return to tasklist",
+                        classes: "govuk-button--secondary",
+                        href: referPaths.show({ referralId: referralId })
+                    }) }}
+                </div>
+            </form>
+        </div>
+
+    </div>
```
yndajas, line 102: We tend to avoid extra line breaks like this one

server/utils/referralUtils.ts
```diff
@@ -72,7 +71,11 @@ export default class ReferralUtils {
       {
         heading: 'Check answers and submit',
         items: [
-          { statusTag: ReferralUtils.taskListStatus('cannot start yet'), text: 'Check answers and submit', url: '#' },
+          {
+            statusTag: ReferralUtils.taskListStatus('cannot start yet'),
+            text: 'Check answers and submit',
+            url: `${referral.id}/check-answers`,
```
yndajas, line 77: Best to generate this using the path from `server/paths` in case it changes

yndajas (commented):
Looking great! Just a few comments on small details and some things we can pick up separately/later

integration_tests/e2eReferDisabled/refer.cy.ts
```diff
@@ -86,4 +86,23 @@ context('Refer', () => {
...
+    cy.signIn()
+
+    const course = courseFactory.build()
+    const courseOffering = courseOfferingFactory.build()
+    const prison = prisonFactory.build({ prisonId: courseOffering.organisationId })
+    const referral = referralFactory.build()
+
+    cy.task('stubCourseByOffering', { course, courseOfferingId: courseOffering.id })
+    cy.task('stubCourseOffering', { courseId: course.id, courseOffering })
+    cy.task('stubPrison', prison)
```
yndajas, line 100: I think the prisoner and referral will need stubbing here as well since those endpoints would be hit by the relevant controller action - we want to be certain we aren't getting a 404 because those resources aren't found in the API vs the page not existing in the UI

server/controllers/refer/referralsController.ts
```diff
@@ -14,6 +14,50 @@ export default class ReferralsController {
...
+
+      res.render('referrals/checkAnswers', {
+        applicationSummaryListRows: ReferralUtils.applicationSummaryListRows(
+          courseOffering,
+          coursePresenter,
+          organisation,
+          person,
+          username,
+        ),
+        courseOfferingId: courseOffering.id,
```
yndajas, line 52: Do we need to pass `courseOfferingId` down? Where is it used?

server/views/referrals/checkAnswers.njk
```diff
@@ -0,0 +1,89 @@
...
+
+{% block personBanner %}
+    {% include "./_personBanner.njk" %}
+{% endblock personBanner %}
+
+{% block backLink %}
+    {{ govukBackLink({
+    text: "Back",
+    href: "#"
+  }) }}
```
yndajas, line 16: This can go to the task list: `referPaths.show({ referralId: referralId })`

server/utils/referralUtils.test.ts
```diff
@@ -101,7 +102,7 @@ describe('ReferralUtils', () => {
                 text: 'cannot start yet',
               },
               text: 'Check answers and submit',
-              url: '#',
+              url: referPaths.checkAnswers({ referralId: referral.id }),
```
yndajas, line 105: Let's (mostly) hardcode this in the test but still use the path method in the utils itself. So here: `/referrals/${referral.id}/checkAnswers` (we tend to do this in tests on code that use path methods)

yndajas (commented):
Looking good - just a few small things

It might also be good to move the integration tests from the third commit into a separate fourth commit (we tend to add new integration tests separate to the code) to keep the commit a bit smaller, but it's not a problem to keep them together if you prefer

server/controllers/refer/referralsController.ts
jsrobertson: Good spot, was used in a previous commit but no longer needed.

integration_tests/e2eReferDisabled/refer.cy.ts
```diff
@@ -86,4 +86,26 @@ context('Refer', () => {
...
+    const courseOffering = courseOfferingFactory.build()
+    const prison = prisonFactory.build({ prisonId: courseOffering.organisationId })
+    const prisoner = prisonerFactory.build()
+    const referral = referralFactory.build()
+
+    cy.task('stubCourseByOffering', { course, courseOfferingId: courseOffering.id })
+    cy.task('stubCourseOffering', { courseId: course.id, courseOffering })
+    cy.task('stubPrison', prison)
+    cy.task('stubPrisoner', prisoner)
+    cy.task('stubReferral', referral)
```
yndajas, line 103: The referral data will need to match the offering and person/prisoner to make sure all the stubs are set up for this specific request: `const referral = referralFactory.build({ offeringId: courseOffering.id, prisonNumber: person.prisonNumber })`

integration_tests/mockApis/auth.ts
```diff
@@ -153,6 +153,8 @@ const stubUserRoles = () =>
     },
   })
 
+export const mockedUsername = 'USER1'
+
```
yndajas, line 157: This could probably go in the default `export` below (and be called with `auth.mockedUsername`, but I'm not too bothered either way

yndajas (approved):
Just one more thing that I think needs fixing, and one suggestion, but I think this is otherwise ready to go

PR #184, opened 2023-09-08
(798) Display person not found error

yndajas:
## Release checklist

### Pre-merge

### Post-merge


server/controllers/refer/referralsController.test.ts
```diff
@@ -92,6 +98,29 @@ describe('ReferralsController', () => {
         courseOfferingId,
         pageHeading: "Enter the person's identifier",
       })
+
+      expect(FormUtils.setFieldErrors).toHaveBeenCalledWith(request, response, ['prisonNumber'])
+      expect(response.locals.errors).toEqual(emptyErrorsLocal)
```
Gweaton, line 103: Let's remove this expectation here, as we're testing that the `response.locals.errors` are set in the `FormUtils` tests.

server/controllers/refer/referralsController.test.ts
```diff
@@ -92,6 +98,29 @@ describe('ReferralsController', () => {
...
+    })
+
+    describe('when there are form errors', () => {
```
Gweaton, line 106: Let's remove this test from the controller, as it's more a test of `setFieldErrors`'s functionality. I think all that's important to test here is that the (mock) function is called with the correct arguments (which we do above). We can test this behaviour in our integration tests.

This should also reduce some of the overhead when adding tests for errors in future.

integration_tests/pages/page.ts
```diff
@@ -165,6 +165,28 @@ export default abstract class Page {
     })
   }
 
+  shouldHaveErrors(errors: Array<{ field: string; message: string }>): void {
```
Gweaton, line 168: I like this 👍

server/controllers/refer/referralsController.test.ts
yndajas: Agree - I did pause on these two tests wondering which should do what/whether either should do some of these assertions

integration_tests/pages/page.ts
yndajas: The error summary bit is inspired by an earlier draft of John's PR

Gweaton (approved):

PR #193, opened 2023-09-12
(653) Reason for referral

jsrobertson:
## Context

https://trello.com/c/LSccMbIN/653-add-reason-for-the-referral-form-step-ui

## Changes in this PR
- Added page template for referral reason form step
- Added `/reason-for-referral` path
- Added `reason` method to `ReferralsController`
- Update `update` method to handle `reason`
- Add url to link to new page from task list

## Screenshots of UI changes
localhost_3000_referrals_778ad55f-aa95-48b7-9f4b-96a58c068c7e_reason

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/integration_tests/e2e/refer.cy.ts
+++ b/integration_tests/e2e/refer.cy.ts
@@ -20,6 +20,7 @@ import {
   TaskListPage,
 } from '../pages/refer'
 import CompletePage from '../pages/refer/complete'
+import ReasonPage from '../pages/refer/reason'
 
 context('Refer', () => {
   beforeEach(() => {
@@ -300,6 +301,77 @@ context('Refer', () => {
 
     const taskListPage = Page.verifyOnPage(TaskListPage, { course, courseOffering, organisation, referral })
     taskListPage.shouldHaveConfirmedOasys()
+  })
+
+  it('Shows the reason form page', () => {
+    cy.signIn()
+
+    const prisoner = prisonerFactory.build({
+      firstName: 'Del',
+      lastName: 'Hatton',
+    })
+    const person = personFactory.build({
+      currentPrison: prisoner.prisonName,
+      name: 'Del Hatton',
+      prisonNumber: prisoner.prisonerNumber,
+    })
+
+    const referral = referralFactory.build({ prisonNumber: person.prisonNumber, reason: undefined })
+
+    cy.task('stubPrisoner', prisoner)
+    cy.task('stubReferral', referral)
+
+    const path = referPaths.reason({ referralId: referral.id })
+    cy.visit(path)
+
+    const reasonPage = Page.verifyOnPage(ReasonPage, { referral })
+    reasonPage.shouldContainNavigation(path)
+    reasonPage.shouldContainBackLink(referPaths.show({ referralId: referral.id }))
+    reasonPage.shouldContainInformationTypeDetails()
+    reasonPage.shouldContainReasonTextArea()
+    reasonPage.shouldContainSaveAndContinueButton()
+  })
+
+  it('On submitting a reason, updates the referral and redirects to the task list', () => {
+    cy.signIn()
+
+    const course = courseFactory.build()
+    const courseOffering = courseOfferingFactory.build()
+
+    const prisoner = prisonerFactory.build({
+      firstName: 'Del',
+      lastName: 'Hatton',
+    })
+    const person = personFactory.build({
+      currentPrison: prisoner.prisonName,
+      name: 'Del Hatton',
+      prisonNumber: prisoner.prisonerNumber,
+    })
+
+    const prison = prisonFactory.build({ prisonId: courseOffering.organisationId })
+    const organisation = OrganisationUtils.organisationFromPrison('an-ID', prison)
+
+    const referral = referralFactory.build({
+      offeringId: courseOffering.id,
+      prisonNumber: person.prisonNumber,
+      reason: undefined,
+    })
+
+    cy.task('stubCourseByOffering', { course, courseOfferingId: courseOffering.id })
+    cy.task('stubCourseOffering', { courseId: course.id, courseOffering })
+    cy.task('stubPrison', prison)
+    cy.task('stubPrisoner', prisoner)
+    cy.task('stubReferral', referral)
+    cy.task('stubUpdateReferral', referral.id)
+
+    const path = referPaths.reason({ referralId: referral.id })
+    cy.visit(path)
+
+    const reasonPage = Page.verifyOnPage(ReasonPage, { referral })
+    reasonPage.submitReason()
+
+    const taskListPage = Page.verifyOnPage(TaskListPage, { course, courseOffering, organisation, referral })
+    taskListPage.shouldHaveReason()
   })
 
   it('Links to the check answers page when the referral is ready for submission', () => {
--- a/integration_tests/pages/refer/reason.ts
+++ b/integration_tests/pages/refer/reason.ts
@@ -1,0 +1,43 @@
+import Page from '../page'
+import type { Referral } from '@accredited-programmes/models'
+
+export default class ReasonPage extends Page {
+  referral: Referral
+
+  constructor(args: { referral: Referral }) {
+    super('Add reason for referral and supporting information')
+
+    const { referral } = args
+
+    this.referral = referral
+  }
+
+  shouldContainInformationTypeDetails() {
+    cy.get('[data-testid="information-type-details"]').then(informationTypeDetailsElement => {
+      this.shouldContainDetails(
+        'What type of information can I add?',
+        "you can provide information about why you're referring this person to this programme" +
+          'you can add supporting information that would be useful for the programme team to know' +
+          'you can add any other information that would be useful for the programme team to know',
+        informationTypeDetailsElement,
+      )
+    })
+  }
+
+  shouldContainReasonTextArea() {
+    this.shouldContainTextArea('reason', 'Add reason for referral and supporting information')
+  }
+
+  shouldContainSaveAndContinueButton() {
+    this.shouldContainButton('Save and continue')
+  }
+
+  submitReason() {
+    const reason = 'This is the reason for the referral'
+    cy.get('[data-testid="reason-text-area"]').type(reason)
+    this.referral = { ...this.referral, reason }
+    // We're stubbing the referral here to make sure the updated referral is available on the task list page
+    cy.task('stubReferral', this.referral)
+    this.shouldContainButton('Save and continue').click()
+  }
+}
--- a/server/@types/models/Referral.ts
+++ b/server/@types/models/Referral.ts
@@ -6,6 +6,7 @@ type Referral = {
   oasysConfirmed: boolean
   offeringId: CourseOffering['id']
   prisonNumber: Person['prisonNumber']
+  reason: string
   referrerId: Express.User['userId']
   status: ReferralStatus
 }
--- a/server/controllers/refer/referralsController.test.ts
+++ b/server/controllers/refer/referralsController.test.ts
@@ -397,7 +397,28 @@ describe('ReferralsController', () => {
         await requestHandler(request, response, next)
 
         expect(response.redirect).toHaveBeenCalledWith(referPaths.show({ referralId: referral.id }))
-        expect(referralService.updateReferral).toHaveBeenCalledWith(token, referral.id, { oasysConfirmed: true })
+        expect(referralService.updateReferral).toHaveBeenCalledWith(token, referral.id, {
+          oasysConfirmed: true,
+          reason: referral.reason,
+        })
+      })
+    })
+
+    describe('updating `reason`', () => {
+      it('asks the service to update the field and redirects to the show action', async () => {
+        const referral = referralFactory.build({ oasysConfirmed: true, reason: undefined })
+        referralService.getReferral.mockResolvedValue(referral)
+
+        request.body.reason = ' Some reason\nAnother paragraph\n '
+
+        const requestHandler = referralsController.update()
+        await requestHandler(request, response, next)
+
+        expect(response.redirect).toHaveBeenCalledWith(referPaths.show({ referralId: referral.id }))
+        expect(referralService.updateReferral).toHaveBeenCalledWith(token, referral.id, {
+          oasysConfirmed: true,
+          reason: 'Some reason\nAnother paragraph',
+        })
       })
     })
   })
--- a/server/controllers/refer/referralsController.test.ts
+++ b/server/controllers/refer/referralsController.test.ts
@@ -488,4 +488,38 @@ describe('ReferralsController', () => {
       })
     })
   })
+
+  describe('reason', () => {
+    it('renders the reason for referral page', async () => {
+      const person = personFactory.build()
+      personService.getPerson.mockResolvedValue(person)
+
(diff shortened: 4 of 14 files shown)


server/controllers/refer/referralsController.ts
```diff
@@ -134,6 +134,27 @@ export default class ReferralsController {
...
+  reason(): TypedRequestHandler<Request, Response> {
+    return async (req: Request, res: Response) => {
+      TypeUtils.assertHasUser(req)
+
+      const referral = await this.referralService.getReferral(req.user.token, req.params.referralId)
+      const person = await this.personService.getPerson(req.user.token, referral.prisonNumber)
+
+      if (!person) {
+        throw createError(404, {
+          userMessage: `Person with prison number ${referral.prisonNumber} not found.`,
```
jsrobertson, line 146: Spotted some of the other methods incorrectly(?) refer to `req.params.prisonNumber` which I can update in another commit/PR.

server/controllers/refer/referralsController.ts
```diff
@@ -243,9 +264,11 @@ export default class ReferralsController {
       const referral = await this.referralService.getReferral(req.user.token, req.params.referralId)
       const oasysConfirmed =
         typeof req.body.oasysConfirmed === 'undefined' ? referral.oasysConfirmed : req.body.oasysConfirmed
+      const reason = typeof req.body.reason === 'undefined' ? referral.reason : req.body.reason
```
jsrobertson, line 267: Might make sense to `trim` the reason to prevent any unwanted whitespace?

server/controllers/refer/referralsController.ts
yndajas: Good catch. I think these have been missed because we aren't testing the message - only the status code. We tried before, but the assertion would pass regardless of what string you pass in

server/controllers/refer/referralsController.ts
yndajas: Yeah, and hopefully to stop up accepting a string like `\n\n\n` becing a truthy `reason` (haven't checked what logic you have on the task list to say whether this is completed yet)

server/views/referrals/reason.njk
```diff
@@ -0,0 +1,59 @@
...
+
+      <form action="{{ referPaths.update({ referralId: referral.id }) }}?_method=PUT" method="post">
+        <input type="hidden" name="_csrf" value="{{ csrfToken }}"/>
+
+        {{ govukTextarea({
+          name: "reason",
+          id: "reason",
+          label: {
+            text: "Referral reason and supporting information"
+          },
```
yndajas, line 49: This isn't in the design, but I think we do need a visible label of some kind, so it makes sense to include it here. I wonder if it would be better to follow the name in the heading here to make it read a bit softer: "Reason for referral and any additional information"

server/controllers/refer/referralsController.test.ts
```diff
@@ -388,4 +388,23 @@ describe('ReferralsController', () => {
...
+      const requestHandler = referralsController.reason()
+      await requestHandler(request, response, next)
+
+      expect(response.render).toHaveBeenCalledWith('referrals/reason', {
+        pageHeading: 'Add reason for referral and supporting information',
+        person,
+        referral,
+      })
+    })
+  })
```
yndajas, line 427: I missed this when reviewing other pages, but we should have a ``describe('when the person service returns `null`'`` block here

server/paths/refer.ts
```diff
@@ -17,6 +17,7 @@ const confirmOasysPath = showReferralPath.path('oasys-confirmed')
 const checkAnswersPath = showReferralPath.path('check-answers')
 const completeReferralPath = showReferralPath.path('complete')
 const submitReferralPath = showReferralPath.path('submit')
+const reasonForReferralPath = showReferralPath.path('reason-for-referral')
```
yndajas, line 20: I think this should go after `confirmOasysPath` - currently they're in roughly chronological order

server/routes/refer.ts
```diff
@@ -22,6 +22,7 @@ export default function routes(controllers: Controllers, router: Router): Router
   get(referPaths.checkAnswers.pattern, referralsController.checkAnswers())
   get(referPaths.complete.pattern, referralsController.complete())
   post(referPaths.submit.pattern, referralsController.submit())
+  get(referPaths.reason.pattern, referralsController.reason())
```
yndajas, line 25: Likewise here (this should go after `confirmOasys`)

server/paths/refer.ts
yndajas, line 20: Interested to hear your thoughts on this, but I used `oasys-confirmed` for the `confirmOasysPath` to have it correspond with the property name (feels more RESTful to me), so we're looking at a referral `/referrals/:referralId` and from that its `/oasysConfirmed` property. Here I wonder if `/reason` would offer greater parity between entities and routes (and actions)?

server/testutils/factories/referral.ts
```diff
@@ -8,6 +8,7 @@ export default Factory.define<Referral>(() => ({
   oasysConfirmed: false,
   offeringId: faker.string.uuid(),
   prisonNumber: faker.string.alphanumeric({ length: 7 }),
+  reason: faker.lorem.sentence(),
```
yndajas, line 11: I saw the conversation about `status` on another PR, but I wonder whether the factory default should be a freshly created referral, pushing us to be intentional when using factories to create more complete referrals, and avoid randomness with values that have functional implications

`oasysConfirmed` defaults to `false` here, not a random boolean, for example

Unsure what the default value is based on the migration here (I guess `''` or `undefined` once it hits our service): https://github.com/ministryofjustice/hmpps-accredited-programmes-api/blob/main/src/main/resources/db/migration/V11__extend_referral_table.sql

server/@types/models/Referral.ts
```diff
@@ -6,6 +6,7 @@ type Referral = {
   oasysConfirmed: boolean
   offeringId: CourseOffering['id']
   prisonNumber: Person['prisonNumber']
+  reason: string
```
yndajas, line 9: Technically the Nunjucks doesn't need to typecheck, but I wonder if this should go before it's used in the template (currently the first commit)

yndajas, line 9: Can we update `ReferralUtils.isReadyForSubmission` and relevant tests with this change?

server/controllers/refer/referralsController.ts
yndajas: Yeah, and hopefully to prevent us reading a string like `\n\n\n` as a truthy `reason` (haven't checked what logic you have on the task list to say whether this is completed yet)

server/controllers/refer/referralsController.test.ts
```diff
@@ -309,7 +309,7 @@ describe('ReferralsController', () => {
   describe('update', () => {
     describe('updating `oasysConfirmed`', () => {
       it('asks the service to update the field and redirects to the show action', async () => {
-        const referral = referralFactory.build({ oasysConfirmed: false })
+        const referral = referralFactory.build({ oasysConfirmed: false, reason: undefined })
```
yndajas, line 312: Does the line at the end of this block not need updating too? Seems to be passing 🤔 Could be a quirk of `undefined` - not defining it === `undefined`

```ts
expect(referralService.updateReferral).toHaveBeenCalledWith(token, referral.id, { oasysConfirmed: true })
```

server/utils/referralUtils.test.ts
```diff
@@ -60,7 +60,7 @@ describe('ReferralUtils', () => {
 
   describe('taskListSections', () => {
     it('returns task list sections for a given referral', () => {
-      const referral = referralFactory.build()
+      const referral = referralFactory.build({ reason: undefined })
```
yndajas, line 63: Per earlier comment, I think this should be the default for the factory, so if that changes, this change can probably be reverted

server/utils/referralUtils.test.ts
```diff
@@ -91,7 +91,11 @@ describe('ReferralUtils', () => {
               url: `/referrals/${referral.id}/oasys-confirmed`,
             },
             {
-              statusTag: { classes: 'govuk-tag--grey moj-task-list__task-completed', text: 'not started' },
+              statusTag: {
+                attributes: { 'data-testid': 'referral-reason-tag' },
```
yndajas, line 95: Everything is `referral` so I think this can just be `reason-tag`? There shouldn't be a conflict on the task list page

integration_tests/pages/refer/reason.ts
```diff
@@ -0,0 +1,43 @@
...
+    this.shouldContainTextArea('reason', 'Referral reason and supporting information')
+  }
+
+  shouldContainSaveAndContinueButton() {
+    this.shouldContainButton('Save and continue')
+  }
+
+  submitReason() {
+    const reason = 'This is the reason for the referral'
+    cy.get('[name="reason"]').type(reason)
```
yndajas, line 37: Is it worth it/possible to add use a `data-testid` for this? I know we don't on a similar method elsewhere, but it might be good to phase them in with new PRs

server/controllers/refer/referralsController.ts
yndajas: Good catch. I think these have been missed because we aren't testing the message - only the status code. We tried before, but the assertion would pass regardless of what string you pass in

yndajas (commented):
Looking good. Just a few suggestions

server/paths/refer.ts
jsrobertson: I was in 2 minds, I did go with `/reason` to begin with and then started to think ahead to if there was ever another "reason" for anything during the referral process but at the moment there isn't so I'm happy to go with just `/reason`

server/controllers/refer/referralsController.test.ts
jsrobertson: If it's `undefined`, it doesn't technically need to be specified, however if we think it will avoid confusion then happy to add `reason: undefined` and continue to do so in future.

server/controllers/refer/referralsController.ts
jsrobertson: I think this might have something to do with specifying an object as the second parameter with `userMessage`.  What's the reason for doing it this way? Looking at [https-errors](https://github.com/jshttp/http-errors) readme, it looks like the error should just be passed as the second parameter.

Gweaton: You're right, I think we've got slightly confused on how we should be handling errors in the application. I think we picked up this use of an object with a `userMessage` field in it from a couple of other projects, but they're using that to display the validation error to the user. We're not doing that anywhere.

I think we should refactor either:
- The code to use this userMessage variable
- The way we handle errors to pass a string (easier to test).

I'll add a card to tidy this up.

server/testutils/factories/referral.ts
Gweaton: As I said in that conversation, I don't think this is all that important, as we should be explicitly setting the values we care about in the tests where they're relevant.

I know you updated this in your other PR @yndajas and I approved that but I think that if we want to represent a new, empty referral, we should use a [trait](https://github.com/thoughtbot/fishery#adding-reusable-builders-traits-to-factories) to represent this, rather than assuming the default factory-built object will represent this early stage referral.

server/controllers/refer/referralsController.test.ts
Gweaton: Do we _need_ to specify that `reason` is undefined here on line 312? Is it relevant to this section of the task list?

server/utils/referralUtils.test.ts
Gweaton: Let's use a trait to represent a `referral_started` Referral to be more explicit here. I think we can do that as a separate tidyup PR rather than blocking this one.

Gweaton (approved):
Ynda's done a good review here, I've left a couple of small comments but I don't think there's anything really blocking this PR from my perspective. They're more things we can tidy up in a future PR.

server/controllers/refer/referralsController.ts
jsrobertson: For the sake of this PR I've added a test to check for the error with a string.  Can refactor the others in the future task

server/controllers/refer/referralsController.test.ts
jsrobertson: Thought about it, probably nicer to remove the undefined and just update the test to check that `referralService.updateReferral` is called with the existing reason value.

yndajas: If we're [keeping a pre-filled](https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/pull/193#discussion_r1324149829) `reason` as the factory default (I'm not sure what I think anymore 😅) then I think checking that it's called with the existing value makes sense here

integration_tests/e2e/refer.cy.ts
```diff
@@ -301,6 +302,77 @@ context('Refer', () => {
     taskListPage.shouldHaveConfirmedOasys()
   })
 
+  it('Shows the reason form page', () => {
```
yndajas, line 305: One thing I missed last night: we'll need a corresponding test in `e2eReferDisabled`

server/views/referrals/reason.njk
yndajas: Should the page heading be updated to "supporting" if we're going for this?

server/testutils/factories/referral.ts
yndajas: Shall we revert the `status` change from my PR and make `oasysConfirmed` a random boolean in that case?

Gweaton: I'll add a card to add a trait, we can tackle that as part of that work

server/controllers/refer/referralsController.test.ts
```diff
@@ -318,7 +318,28 @@ describe('ReferralsController', () => {
...
+        request.body.reason = ' Some reason\nAnother paragraph\n '
+
+        const requestHandler = referralsController.update()
+        await requestHandler(request, response, next)
+
+        expect(response.redirect).toHaveBeenCalledWith(referPaths.show({ referralId: referral.id }))
+        expect(referralService.updateReferral).toHaveBeenCalledWith(token, referral.id, {
+          oasysConfirmed: true,
+          reason: 'Some reason\nAnother paragraph',
+        })
```
yndajas, line 342: Shall we do the same as with `oasysConfirmed` here - not set `oasysConfirmed` when building the referral, but check that the current value is sent to `updateReferral`?

server/controllers/refer/referralsController.test.ts
yndajas: Related: https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/pull/193/files#r1324300112

server/views/referrals/reason.njk
Gweaton: @yndajas we've had a quick chat about this and have made the label a page title and removed this extra one. I was actually looking at outdated content so it should actually be "additional information"

yndajas: Figma! 😅

server/controllers/refer/referralsController.test.ts
jsrobertson: Think this has highlighted an issue with setting `request` values in each of the tests without resetting.  Will investigate.

Gweaton (approved):
Still to do, in future PRs:

- Make this field required
- Ensure this section has been filled in before being able to Check your answers
- Play this answer back on the CYA page

PR #197, opened 2023-09-13
(650d) Update check your answers page validation

jsrobertson:
## Context

Utilise `setFieldErrors` from `FormUtils` to set the validation for the check your answers page.

## Changes in this PR
- Use `FormUtils.setFieldErrors` for the `confirmation` checkbox on check your answers page
- Updated the referrals controller test to reset the response value between each test to avoid unexpected test behaviour.

## Screenshots of UI changes
localhost_3000_referrals_e38891f9-033b-412c-96f1-60b6689a8109_check-answers

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/integration_tests/pages/page.ts
+++ b/integration_tests/pages/page.ts
@@ -203,8 +203,6 @@ export default abstract class Page {
         const { actual, expected } = Helpers.parseHtml(fieldErrorElement, `Error: ${error.message}`)
         expect(actual).to.equal(expected)
       })
-
-      cy.get(`[name="${error.field}"]`).should('have.class', 'govuk-input--error')
     })
   }
 
--- a/integration_tests/pages/page.ts
+++ b/integration_tests/pages/page.ts
@@ -72,18 +72,6 @@ export default abstract class Page {
     })
   }
 
-  shouldContainErrorSummary(errors: Array<{ href: string; text: string }>): void {
-    cy.get('.govuk-error-summary').should('exist')
-    cy.get('.govuk-error-summary__list').within(() => {
-      cy.get('li').should('have.length', errors.length)
-      cy.get('li').each((errorElement, errorElementIndex) => {
-        const { actual, expected } = Helpers.parseHtml(errorElement, errors[errorElementIndex].text)
-        expect(actual).to.equal(expected)
-        cy.get('a').should('have.attr', 'href', errors[errorElementIndex].href)
-      })
-    })
-  }
-
   shouldContainLink(text: string, href: string): void {
     cy.contains('.govuk-link', text).should('have.attr', 'href', href)
   }
--- a/server/controllers/refer/referralsController.test.ts
+++ b/server/controllers/refer/referralsController.test.ts
@@ -266,9 +266,6 @@ describe('ReferralsController', () => {
 
   describe('checkAnswers', () => {
     it('renders the referral check answers page', async () => {
-      const errors: Array<string> = []
-      request.flash = jest.fn().mockReturnValue(errors)
-
       const person = personFactory.build()
       personService.getPerson.mockResolvedValue(person)
 
@@ -296,6 +293,11 @@ describe('ReferralsController', () => {
       const requestHandler = referralsController.checkAnswers()
       await requestHandler(request, response, next)
 
+      const emptyErrorsLocal = { list: [], messages: {} }
+      ;(FormUtils.setFieldErrors as jest.Mock).mockImplementation((_request, _response, _fields) => {
+        response.locals.errors = emptyErrorsLocal
+      })
+
       expect(response.render).toHaveBeenCalledWith('referrals/checkAnswers', {
         applicationSummaryListRows: ReferralUtils.applicationSummaryListRows(
           courseOffering,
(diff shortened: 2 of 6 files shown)


integration_tests/pages/page.ts
```diff
@@ -203,8 +191,6 @@ export default abstract class Page {
         const { actual, expected } = Helpers.parseHtml(fieldErrorElement, `Error: ${error.message}`)
         expect(actual).to.equal(expected)
       })
-
-      cy.get(`[name="${error.field}"]`).should('have.class', 'govuk-input--error')
```
jsrobertson, line 207: For a checkbox, the error class is on a form group containing the input.

server/controllers/refer/referralsController.test.ts
```diff
@@ -36,8 +33,15 @@ describe('ReferralsController', () => {
   const courseOffering = courseOfferingFactory.build()
 
   let referralsController: ReferralsController
+  let request: DeepMocked<Request>
+  let response: DeepMocked<Response>
+  let next: DeepMocked<NextFunction>
```
jsrobertson, line 38: May not be necessary for `next`

jsrobertson: Also open to alternative methods, but this is how I've done it previously.  Set it as a let and then reset it in the `beforeEach`

server/controllers/refer/referralsController.test.ts
```diff
@@ -36,8 +33,15 @@ describe('ReferralsController', () => {
   const courseOffering = courseOfferingFactory.build()
 
   let referralsController: ReferralsController
+  let request: DeepMocked<Request>
```
jsrobertson, line 36: Just seen you've more or less done the same in #195 @yndajas.  Will wait for that to be merged then rebase.

yndajas: Ah! I see you reset `request` and `next` too - I don't think we ever alter those, but it's probably a good idea in case we do?

server/controllers/refer/referralsController.ts
```diff
@@ -245,12 +245,10 @@ export default class ReferralsController {
...
       if (req.body.confirmation !== 'true') {
-        req.flash('errors', [
-          {
-            href: '#confirmation',
-            text: 'Please confirm that the information you have provided is complete, accurate and up to date',
-          } as unknown as string,
-        ])
+        req.flash(
+          'confirmationError',
+          'Please confirm that the information you have provided is complete, accurate and up to date',
```
Gweaton, line 250: Let's remove "Please" from here - [according to](https://design-system.service.gov.uk/components/error-message/#be-clear-and-concise) the GOV.UK design system it "implies the user has a choice" 😁

Gweaton (approved):
Looks good to me, just one thing to address before merging

server/controllers/refer/referralsController.test.ts
```diff
@@ -24,7 +24,7 @@ jest.mock('../../utils/referralUtils')
 describe('ReferralsController', () => {
   const token = 'SOME_TOKEN'
   let request: DeepMocked<Request>
-  const response: DeepMocked<Response> = createMock<Response>({})
+  let response: DeepMocked<Response>
```
yndajas, line 27: Could we do this in the other controller test files too?

server/controllers/refer/referralsController.test.ts
```diff
@@ -265,9 +266,6 @@ describe('ReferralsController', () => {
 
   describe('checkAnswers', () => {
     it('renders the referral check answers page', async () => {
-      const errors: Array<string> = []
-      request.flash = jest.fn().mockReturnValue(errors)
```
yndajas, line 269: Are we happy with the call to `flash` being unmocked?

yndajas (approved):
Looks good, just a couple of queries

server/controllers/refer/referralsController.test.ts
yndajas: Commit message here has a typo: `referalls` -> `referrals`

server/controllers/refer/referralsController.test.ts
yndajas: This was meant in reference to the final commit - does the flash not need mocking there? Perhaps not

server/controllers/refer/referralsController.ts
yndajas: "Confirm that" sounds a little harsh though - does it need further revision? "I confirm that"?

integration_tests/pages/page.ts
yndajas: Just seen this. Can you mention this in the commit message since there's a change in method functionality (not just use) here?

I suppose the bit above is testing that the field error is present, so this should be okay, unless we can come up with a generic way to test that the error class is present

server/controllers/refer/referralsController.test.ts
yndajas: (I suppose it makes no difference to the outcome of the action in those conditions, but it's still hit)

server/controllers/refer/referralsController.ts
Gweaton: I think it's ok to be direct, I don't think it's too harsh!

server/controllers/refer/referralsController.test.ts
jsrobertson: `.flash` should already be mocked as a `jest.fn()` as a result of `request = createMock<Request>({ user: { token } })` so it doesn't need it here as well.

PR #199, opened 2023-09-15
(842) Reason validation

jsrobertson:
## Context

https://trello.com/c/OkMpaWcU/842-make-reason-for-referral-required-ui

We have decided to create a new controller for each step of a referral to prevent the possibility of the `ReferralsController` becoming too unwieldy.

## Changes in this PR
This extracts the logic for showing, setting and updating the `reason` field on a Referral out of the ReferralsController and into an `ReasonController`.

- Add new `ReasonController`
- Move reason related method and functionality from ReferralsController to new `ReasonController`
- Add new `update` reason method 
- Display error messaging on reason page when no reason submitted

## Screenshots of UI changes

localhost_3000_referrals_a16f58c0-1c2a-495e-ae8d-823e5e9518c5_reason

## Release checklist

[Release process documentation](../doc/how-to/perform-a-release.md)

As part of our continuous deployment strategy we must ensure that this work is
ready to be released once merged.

### Pre-merge

- [ ] There are changes required to the Accredited Programmes API for this change to work...
  - [ ] ... and they have been released to production already

### Post-merge

- [ ] This adds, extends or meaningfully modifies a feature...
  - [ ] ... and I have written or updated an end-to-end test for the happy path in the [Accredited Programmes E2E repo](https://github.com/ministryofjustice/hmpps-accredited-programmes-e2e)
- [ ] This makes new expectations of the API...
  - [ ] ... and I have notified the API developer(s) of changes to the contract tests (Pact), or the API is already compliant
- [ ] [Manually approve](../doc/how-to/perform-a-release.md#releasing-to-the-preprod-environment) release to preprod
- [ ] [Manually approve](../doc/how-to/perform-a-release.md#releasing-to-the-production-environment) release to prod

PR #200, opened 2023-09-18
(850) Refactor controller for confirming OASys information

Gweaton:
## Context

https://trello.com/c/bBzawaKi/850-refactor-oasys-confirmation-out-of-referralscontroller-ui

Now that we’re adding validations to the service e.g. the user must select this checkbox, we’ll need to have an individual `update` function for each field on the Referral for checking each of these fields. We could have added multiple functions to the `ReferralsController` (one for each field) but this would cause the ReferralsController to become a very long file, and a class with multiple responsibilities that looks in danger of becoming unwieldy as the Referral form grows.

## Changes in this PR

This extracts the logic for showing, setting and updating the `oasysConfirmed` field on a Referral out of the `ReferralsController` and into an `OasysConfirmationController` for the reasons mentioned above.

It also renames the URL from `oasys-confirmed` (after the field) to `confirm-oasis` (slightly more “action”-oriented), which I think will make more sense to the end user.

This PR introduces what may look like inconsistencies with regards to naming:
- `oasysConfirmationController`: the controller used for rendering the page and setting the field.
- `confirmOasys`: the action we refer to for setting the `oasysConfirmed` field.
- `oasysConfirmed`: the field on the referral

However, I can’t foresee this causing any confusion for future developers of the service, as I think the actions make sense.

This refactoring will make it easier for us to introduce validations on this checkbox (up next).

N.B: I've left the tests and functionality for the `ReferralController`'s `update` method in place here, as I didn't want to remove them until we'd moved the `reason`-setting behaviour into its own controller, but they will be removed in future.

## Release checklist

### Pre-merge

### Post-merge


server/controllers/refer/oasysConfirmationController.test.ts
```diff
@@ -0,0 +1,90 @@
+import type { DeepMocked } from '@golevelup/ts-jest'
+import { createMock } from '@golevelup/ts-jest'
+import type { NextFunction, Request, Response } from 'express'
+import createError from 'http-errors'
+
+import OasysConfirmationController from './oasysConfirmationController'
+import { referPaths } from '../../paths'
+import type PersonService from '../../services/personService'
+import type ReferralService from '../../services/referralService'
```
yndajas, line 9: Index

server/controllers/refer/oasysConfirmationController.ts
```diff
@@ -0,0 +1,54 @@
+import type { Request, Response, TypedRequestHandler } from 'express'
+import createError from 'http-errors'
+
+import type { ReferralUpdate } from '../../@types/models/Referral'
```
yndajas, line 4: Alias

server/controllers/refer/oasysConfirmationController.ts
```diff
@@ -0,0 +1,54 @@
+import type { Request, Response, TypedRequestHandler } from 'express'
```
yndajas, line 1: The file and class names aren't plural but I guess that's okay? Same with the view

server/views/referrals/oasysConfirmation/show.njk
```diff
@@ -0,0 +1,70 @@
...
+      {% endset %}
+
+      {{ govukNotificationBanner({
+        html: notificationBannerHtml,
+        attributes: {
+          "data-testid": "last-updated-notification-banner"
+        }
+      }) }}
+
+      <form action="{{ referPaths.update({ referralId: referral.id }) }}?_method=PUT" method="post">
```
yndajas: Should this be the new controller action (with a path added)? (in commit one)

server/paths/refer.ts
```diff
@@ -13,7 +13,7 @@ const personPath = peoplePathBase.path(':prisonNumber')
 const referralsPath = path('/referrals')
 const showReferralPath = referralsPath.path(':referralId')
 const referralPersonPath = showReferralPath.path('person')
-const confirmOasysPath = showReferralPath.path('oasys-confirmed')
+const confirmOasysPath = showReferralPath.path('confirm-oasys')
```
yndajas, line 16: I'm a little unsure about this. I suppose `check-answers` does relate to a field as well (`status`) and doesn't use that field as its path, but it has a separate update status endpoint, whereas this is still submitting to the standard referral update path

I suppose it does read better

I wonder if it's worth clarifying in the commit message that this is for the more action-y pages, where you're either doing something or not (confirming/submitting or not), rather than adding data that can take many different forms? Otherwise I wonder if the commit could read as moving towards action-oriented paths for all referral form pages (despite the mention of the reason distinction)

server/controllers/refer/referralsController.test.ts
```diff
@@ -236,44 +236,6 @@ describe('ReferralsController', () => {
     })
```
yndajas, line 236: I think the `update` action needs updating in this (second) commit too (to remove the `oasysConfirmation` handling, and remove the related test)

yndajas (commented):
Can't remember what you said you were going to change here, but here's a few things I spotted from a quick look

server/views/referrals/oasysConfirmation/show.njk
Gweaton: Yep, this is the thing I forgot to do 🤦

yndajas: Oops

server/controllers/refer/referralsController.test.ts
Gweaton: That's going to come in a future PR:
> N.B: I've left the tests and functionality for the ReferralController's update method in place here, as I didn't want to remove them until we'd moved the reason-setting behaviour into its own controller, but they will be removed in future.

server/controllers/refer/oasysConfirmationController.ts
Gweaton: Although I tend to prefer using plurals here, something about `oasysConfirmations` feels wrong to me. Going to keep it singular for now, and I think it lines up ok with the `reasonController` in #199

server/paths/refer.ts
Gweaton: Thanks for your thoughts, it's a bit of a grey area for sure, but I've updated the commit message to hopefully clarify this.

server/routes/refer.ts
```diff
@@ -18,7 +18,8 @@ export default function routes(controllers: Controllers, router: Router): Router
   get(referPaths.show.pattern, referralsController.show())
   put(referPaths.update.pattern, referralsController.update())
   get(referPaths.showPerson.pattern, referralsController.showPerson())
-  get(referPaths.confirmOasys.pattern, referralsController.confirmOasys())
+  get(referPaths.confirmOasys.pattern, oasysConfirmationController.show())
+  post(referPaths.confirmOasys.pattern, oasysConfirmationController.update())
```
yndajas: It looks like the `post` here is added but unused, then changed to `put` later?

server/controllers/refer/referralsController.test.ts
yndajas: Hmm okay. I was thinking something like [this from the reason PR](https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/pull/199/commits/05f100011e38703149cff0c67a8a19717a522fea) since the `oasysConfirmed` update functionality is no longer used with these changes, but it could wait for a separate cleanup PR

yndajas (approved):
Just a couple of things, but nothing blocking

server/controllers/refer/referralsController.test.ts
Gweaton: I suppose depending on who's PR is merged first we can remove this in the later branch

PR #218, opened 2023-09-26
(655b) Update course client and service for creating a participation

jsrobertson:
## Context

https://trello.com/c/i2gzZOoW/655-add-page-for-selecting-course-name-ui

## Changes in this PR
Adds new methods to the course client and course service to create a new course participation.

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/server/data/courseClient.test.ts
+++ b/server/data/courseClient.test.ts
@@ -5,6 +5,7 @@ import CourseClient from './courseClient'
 import config from '../config'
 import { apiPaths } from '../paths'
 import { courseFactory, courseOfferingFactory, courseParticipationFactory, personFactory } from '../testutils/factories'
+import type { CourseParticipation } from '@accredited-programmes/models'
 
 pactWith({ consumer: 'Accredited Programmes UI', provider: 'Accredited Programmes API' }, provider => {
   let courseClient: CourseClient
@@ -64,6 +65,68 @@ pactWith({ consumer: 'Accredited Programmes UI', provider: 'Accredited Programme
       const result = await courseClient.all()
 
       expect(result).toEqual(allCourses)
+    })
+  })
+
+  describe('createParticipation', () => {
+    describe('when `courseId` is provided', () => {
+      const courseParticipation = courseParticipationFactory.withCourseId().build()
+      const { courseId, prisonNumber } = courseParticipation
+
+      beforeEach(() => {
+        provider.addInteraction({
+          state: `A person exists with prison number ${prisonNumber}`,
+          uponReceiving: `A request to create a course participation for person "${prisonNumber}"`,
+          willRespondWith: {
+            body: Matchers.like(courseParticipation),
+            status: 201,
+          },
+          withRequest: {
+            body: { courseId, prisonNumber } as CourseParticipation & { courseId: string },
+            headers: {
+              authorization: `Bearer ${token}`,
+            },
+            method: 'POST',
+            path: apiPaths.participations.create({}),
+          },
+        })
+      })
+
+      it('creates a course participation using the `courseId` value', async () => {
+        const result = await courseClient.createParticipation(prisonNumber, courseId)
+
+        expect(result).toEqual(courseParticipation)
+      })
+    })
+
+    describe('when `otherCourseName` is provided and `courseId` is not', () => {
+      const courseParticipation = courseParticipationFactory.withOtherCourseName().build()
+      const { otherCourseName, prisonNumber } = courseParticipation
+
+      beforeEach(() => {
+        provider.addInteraction({
+          state: `A person exists with prison number ${prisonNumber}`,
+          uponReceiving: `A request to create a course participation for person "${prisonNumber}"`,
+          willRespondWith: {
+            body: Matchers.like(courseParticipation),
+            status: 201,
+          },
+          withRequest: {
+            body: { otherCourseName, prisonNumber } as CourseParticipation & { otherCourseName: string },
+            headers: {
+              authorization: `Bearer ${token}`,
+            },
+            method: 'POST',
+            path: apiPaths.participations.create({}),
+          },
+        })
+      })
+
+      it('creates a course participation using the `otherCourseName` value', async () => {
+        const result = await courseClient.createParticipation(prisonNumber, undefined, otherCourseName)
+
+        expect(result).toEqual(courseParticipation)
+      })
     })
   })
 
--- a/server/data/courseClient.ts
+++ b/server/data/courseClient.ts
@@ -15,6 +15,17 @@ export default class CourseClient {
     return (await this.restClient.get({ path: apiPaths.courses.index({}) })) as Array<Course>
   }
 
+  async createParticipation(
+    prisonNumber: CourseParticipation['prisonNumber'],
+    courseId?: CourseParticipation['id'],
+    otherCourseName?: CourseParticipation['otherCourseName'],
+  ): Promise<CourseParticipation> {
+    return (await this.restClient.post({
+      data: { courseId, otherCourseName, prisonNumber },
+      path: apiPaths.participations.create({}),
+    })) as CourseParticipation
+  }
+
   async find(courseId: Course['id']): Promise<Course> {
     return (await this.restClient.get({ path: apiPaths.courses.show({ courseId }) })) as Course
   }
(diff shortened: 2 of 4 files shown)


server/data/courseClient.ts
```diff
@@ -15,6 +15,17 @@ export default class CourseClient {
...
 
+  async createParticipation(
+    prisonNumber: Person['prisonNumber'],
+    courseId?: Course['id'],
+    otherCourseName?: string,
+  ): Promise<CourseParticipation> {
+    return this.restClient.post({
+      data: { courseId, otherCourseName, prisonNumber },
+      path: apiPaths.participations.create({}),
+    }) as Promise<CourseParticipation>
```
yndajas, line 26: I think the `as` doesn't need the `Promise`?

server/data/courseClient.ts
```diff
@@ -15,6 +15,17 @@ export default class CourseClient {
     return (await this.restClient.get({ path: apiPaths.courses.index({}) })) as Array<Course>
   }
 
+  async createParticipation(
+    prisonNumber: Person['prisonNumber'],
+    courseId?: Course['id'],
+    otherCourseName?: string,
```
yndajas, line 21: I think these types can all come from `CourseParticipation`

server/data/courseClient.test.ts
```diff
@@ -67,6 +67,37 @@ pactWith({ consumer: 'Accredited Programmes UI', provider: 'Accredited Programme
...
+    beforeEach(() => {
+      provider.addInteraction({
+        state: `A person exists with prison number ${prisonNumber}`,
+        uponReceiving: `A request to create a course participation for person "${prisonNumber}"`,
+        willRespondWith: {
+          body: Matchers.like(courseParticipations[0]),
+          status: 201,
+        },
+        withRequest: {
+          body: { courseId, otherCourseName, prisonNumber },
```
yndajas, line 84: Should the body not be `courseParticipations[0].courseId` etc?

server/data/courseClient.test.ts
```diff
@@ -67,6 +67,37 @@ pactWith({ consumer: 'Accredited Programmes UI', provider: 'Accredited Programme
...
 
+  describe('createParticipation', () => {
+    const { prisonNumber } = person
+    const courseId = course1.id
+    const otherCourseName = 'Other course'
+
+    beforeEach(() => {
+      provider.addInteraction({
+        state: `A person exists with prison number ${prisonNumber}`,
+        uponReceiving: `A request to create a course participation for person "${prisonNumber}"`,
```
yndajas, line 78: I wonder if we need to have two tests here: one for passing a defined `courseId` and one for passing a defined `otherCourseName` (with the other `undefined` in each case)?

server/services/courseService.ts
```diff
@@ -4,6 +4,22 @@ import type { Course, CourseOffering, CourseParticipation, Person } from '@accre
 export default class CourseService {
   constructor(private readonly courseClientBuilder: RestClientBuilder<CourseClient>) {}
 
+  async createParticipation(
+    token: Express.User['token'],
+    prisonNumber: Person['prisonNumber'],
+    courseId: Course['id'],
+    otherCourseName: string,
```
yndajas, line 11: Types from `CourseParticipation` again

server/services/courseService.ts
```diff
@@ -4,6 +4,22 @@ import type { Course, CourseOffering, CourseParticipation, Person } from '@accre
...
+  ): Promise<CourseParticipation> {
+    const courseClient = this.courseClientBuilder(token)
+    const isOtherCourse = courseId === 'other'
+
+    return courseClient.createParticipation(
+      prisonNumber,
+      isOtherCourse ? undefined : courseId,
+      isOtherCourse ? otherCourseName : undefined,
+    )
```
yndajas, line 20: Interesting! I think this should be either in the service or the controller rather than the client

Is this tied to the specific way the new participations page works? Would there ever be a case where we need to use this service method with a course that has an `otherCourseName` but without having set `courseId` to `'other'`? Perhaps not

server/services/courseService.test.ts
```diff
@@ -19,6 +20,46 @@ describe('CourseService', () => {
     courseClientBuilder.mockReturnValue(courseClient)
   })
 
+  describe('createParticipation', () => {
+    const person = personFactory.build()
+
+    it('creates a course participation', async () => {
```
yndajas, line 26: I wonder if it's worth putting this in a `describe` block too? I'm not sure whether we should treat having a real `courseId` as more default than an `otherCourseName`? Perhaps it is

server/services/courseService.test.ts
```diff
@@ -19,6 +20,46 @@ describe('CourseService', () => {
...
+
+      expect(courseClientBuilder).toHaveBeenCalledWith(token)
+      expect(courseClient.createParticipation).toHaveBeenCalledWith(person.prisonNumber, courseId, undefined)
+    })
+
+    describe('when the courseId is "other"', () => {
+      it('creates a course participation with the other course name', async () => {
+        const courseId = 'other'
+        const otherCourseName = 'Other course'
+        const courseParticipation = courseParticipationFactory.withOtherCourseName().build()
```
yndajas, line 47: It doesn't matter because of the mocking, but I wonder if it's worth passing the `otherCourseName` into build here for tests as documentation purposes? Likewise with `courseId` in the previous test

yndajas (commented):
Looks good. A few small suggestions. I guess this wouldn't break dev until it gets used in the controllers/views?

server/data/courseClient.ts
jsrobertson: Yeah needs it otherwise you get `Conversion of type 'Promise<unknown>' to type 'CourseParticipation' may be a mistake because neither type sufficiently overlaps with the other.`. Did wonder if it might be nice to update the rest client methods to take something like `<T>` so we can specify the response. One for another PR/Spike though.

yndajas: Oh that's strange! The others seem not to wrap the `as` types in promises 🤔

server/services/courseService.ts
jsrobertson: I initially had it in the client but then thought it would be nicer in the service.  However now you mention it, 'other' is linked to the functionality of the page template so might make sense to keep anything related to courseId === 'other' in the controller.

yndajas: Yeah, I don't know if we'll ever need this to work with other pages, but it does feel more like a page-specific expectation

jsrobertson: Moved to controller, see #222

server/data/courseClient.ts
jsrobertson: The others wrap the returned await in `()`.  Will update to be be consistent for now but be interested to investigate if that's necessary.

yndajas: Ah, just seen this after commenting in the other PR! Yeah, perhaps these can all just be promises, like in the services, but best to be consistent unless we're updating the rest

server/data/courseClient.test.ts
```diff
@@ -67,6 +68,68 @@ pactWith({ consumer: 'Accredited Programmes UI', provider: 'Accredited Programme
...
+      beforeEach(() => {
+        provider.addInteraction({
+          state: `A person exists with prison number ${prisonNumber}`,
+          uponReceiving: `A request to create a course participation for person "${prisonNumber}"`,
+          willRespondWith: {
+            body: Matchers.like(courseParticipation),
+            status: 201,
+          },
+          withRequest: {
+            body: { courseId, prisonNumber } as CourseParticipation,
```
yndajas, line 85: Ahhh I see what's going on with the casting here. It looks like the problem is that `courseId` could be `undefined`, which would make it invalid JSON - the `body` here needs to be valid JSON. This isn't a `CourseParticipation` though - that would require `id`, `addedBy` and `createdAt` at a minimum - so it might be better to either cast `courseId` as a `string`, or update L73 to:

```ts
      const courseParticipation = courseParticipationFactory.withCourseId().build() as CourseParticipation & {
        courseId: string
      }
```

That would narrow out the possibility of it being `undefined`

Same in the test below

yndajas (approved):
Just one last suggestion to clarify some of the typing here, but I think it's otherwise good to go!

PR #225, opened 2023-09-28
Update checkAccessibility cypress command to accept optional accessibilityRules

jsrobertson:
## Context

Now we plan to use conditional radio buttons from govuk-frontend, we were getting an accessibility error which were preventing the tests from passing.

## Changes in this PR

Updated the cypress `checkAccessibility` command to accept some rules, which we can then use on a page by page basis.
 
For example in page which has conditional radio buttons we can do:
`super('Add Accredited Programme history', { accessibilityRules: { 'aria-allowed-attr': { enabled: false } } })`

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/integration_tests/support/commands.ts
+++ b/integration_tests/support/commands.ts
@@ -1,3 +1,4 @@
+import type { AxeRules } from '@accredited-programmes/integration-tests'
 import type { Result } from 'axe-core'
 
 Cypress.Commands.add('signIn', (options = { failOnStatusCode: true }) => {
@@ -23,8 +24,8 @@ const logAccessibilityViolations = (violations: Array<Result>) => {
   cy.task('table', violationsData)
 }
 
-Cypress.Commands.add('checkAccessibility', () => {
+Cypress.Commands.add('checkAccessibility', (rules: AxeRules = undefined) => {
   cy.injectAxe()
   cy.configureAxe({ rules: [{ enabled: false, id: 'region', selector: '.govuk-phase-banner' }] })
-  cy.checkA11y(undefined, undefined, logAccessibilityViolations)
+  cy.checkA11y(undefined, { rules }, logAccessibilityViolations)
 })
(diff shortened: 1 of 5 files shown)


integration_tests/tsconfig.json
```diff
@@ -10,6 +10,7 @@
       "@accredited-programmes/models": ["../server/@types/models/index.d.ts"],
       "@accredited-programmes/ui": ["../server/@types/ui/index.d.ts"],
       "@govuk-frontend": ["../server/@types/govukFrontend/index.d.ts"],
+      "@accredited-programmes/integration-tests": ["./@types/index.d.ts"],
```
jsrobertson, line 13: Will nudge this up

integration_tests/support/commands.ts
```diff
@@ -23,8 +24,8 @@ const logAccessibilityViolations = (violations: Array<Result>) => {
   cy.task('table', violationsData)
 }
 
-Cypress.Commands.add('checkAccessibility', () => {
+Cypress.Commands.add('checkAccessibility', (rules: AxeRules = undefined) => {
   cy.injectAxe()
   cy.configureAxe({ rules: [{ enabled: false, id: 'region', selector: '.govuk-phase-banner' }] })
-  cy.checkA11y(undefined, undefined, logAccessibilityViolations)
+  cy.checkA11y(undefined, { rules }, logAccessibilityViolations)
```
Gweaton, line 30: Are we always passing `{ rules: undefined | AxeRules }` here? Is it ok to be passing `{ rules: undefined }` if so? Not saying it's a problem, just curious - if it works, it works!

Gweaton (approved):
I think this looks good to me, just had one question, really.

integration_tests/support/commands.ts
jsrobertson: Yeah there doesn't appear to be any issues with it being undefined.  I'm confident it works as I've got it on another branch with the integration tests for the Select programme page.

PR #228, opened 2023-09-28
Tidy up

yndajas:
## Changes in this PR

Couple of small changes to simplify/DRY up some of our code

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/server/routes/refer.ts
+++ b/server/routes/refer.ts
@@ -6,7 +6,7 @@ import { referPaths } from '../paths'
 import { RouteUtils } from '../utils'
 
 export default function routes(controllers: Controllers, router: Router): Router {
-  const { get, post, put } = RouteUtils.actions(router)
+  const { get, post, put } = RouteUtils.actions(router, { allowedRoles: [ApplicationRoles.ACP_REFERRER] })
   const {
     courseParticipationsController,
     reasonController,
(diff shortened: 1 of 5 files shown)


server/routes/refer.ts
```diff
@@ -6,7 +6,7 @@ import { referPaths } from '../paths'
 import { RouteUtils } from '../utils'
 
 export default function routes(controllers: Controllers, router: Router): Router {
-  const { get, post, put } = RouteUtils.actions(router)
+  const { get, post, put } = RouteUtils.actions(router, { allowedRoles: [ApplicationRoles.ACP_REFERRER] })
```
Gweaton, line 9: Ah this is a good way to make the file a bit more readable, nice work. I think it might be good to document what overwriting these to make them e.g. open to all users would look like in the commit message - would we pass an empty object here or list all of the roles it's available to?

jsrobertson: Like this, nice one.

yndajas: I think it would be exactly the same as before, since the previous and this approach should both resolve to `undefined` if the argument isn't passed in

yndajas: (but yeah we should document it - I'll pick it up on Monday unless anyone wants to tidy this up tomorrow)

Gweaton (approved):

PR #230, opened 2023-09-29
Specify `yearStarted` to prevent intermittent test failure

jsrobertson:
## Context

This test was sometimes failing because the course participation related factories randomly provide a `yearStarted` in `outcome` if not specified in `build`.

## Changes in this PR
Provide a `yearStarted` value to problematic test.  

### Before
image

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/server/utils/courseParticipationUtils.test.ts
+++ b/server/utils/courseParticipationUtils.test.ts
@@ -20,6 +20,7 @@ describe('CourseParticipationUtils', () => {
           detail: 'Motivation level over 9000!',
           status: 'complete',
           yearCompleted: 2019,
+          yearStarted: undefined,
         },
         setting: {
           location: 'Greater Tharfoot',


server/utils/courseParticipationUtils.test.ts
```diff
@@ -20,6 +20,7 @@ describe('CourseParticipationUtils', () => {
           detail: 'Motivation level over 9000!',
           status: 'complete',
           yearCompleted: 2019,
+          yearStarted: 2018,
```
Gweaton, line 23: Ooh, this is an interesting issue. A `courseParticipation` shouldn't have both a `yearStarted` and a `yearCompleted`. If the outcome is "complete" it has `yearCompleted` but if it's "incomplete" it has `yearStarted`. We should probably add some logic to the factory for this, or a couple of traits.

To get this intermittent flakiness to pass though, let's make this `null`

jsrobertson: Ah didn't realise that was the case, cheers.

Gweaton (approved):

server/utils/courseParticipationUtils.test.ts
yndajas: I'm confused about why this is happening. I would've though if you pass in an object with a defined `outcome` it would remove the randomness of that property, but it seems not 🤔

Gweaton: Yeah, I'm surprised about that too...I have a feeling it merges the params under the hood.

yndajas: I wonder if we need to specify `undefined` explicitly on every missing property in the participation and participation outcome traits

PR #232, opened 2023-09-29
(851a) Update course client and service for finding and updating course participations 

jsrobertson:
## Context

https://trello.com/c/oMnAOOLO/851-add-page-for-entering-course-details-ui-m

For the course participation details page we need to make a couple of calls.  One to get the participation and check it exists and then another to update when the form has been filled in.

## Changes in this PR

- Add `CourseParticipationUpdate` type. 
- Add `findParticipation` and `updateParticipation` to `CourseClient`
- Add `getParticipation` and `updateParticipation` to `CourseService`

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/server/@types/models/CourseParticipation.ts
+++ b/server/@types/models/CourseParticipation.ts
@@ -25,8 +25,22 @@ type CourseParticipation = {
   source?: string
 }
 
+type CourseParticipationUpdate = {
+  outcome: CourseParticipationOutcome
+  setting: CourseParticipationSetting
+  courseId?: CourseParticipation['courseId']
+  otherCourseName?: CourseParticipation['otherCourseName']
+  source?: CourseParticipation['source']
+}
+
 type CourseParticipationWithName = CourseParticipation & {
   name: string
 }
 
-export type { CourseParticipation, CourseParticipationOutcome, CourseParticipationSetting, CourseParticipationWithName }
+export type {
+  CourseParticipation,
+  CourseParticipationOutcome,
+  CourseParticipationSetting,
+  CourseParticipationUpdate,
+  CourseParticipationWithName,
+}
--- a/server/services/courseService.test.ts
+++ b/server/services/courseService.test.ts
@@ -1,3 +1,5 @@
+import { faker } from '@faker-js/faker'
+import createError from 'http-errors'
 import { when } from 'jest-when'
 
 import CourseService from './courseService'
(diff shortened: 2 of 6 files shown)


server/@types/models/CourseParticipation.ts
```diff
@@ -25,8 +25,22 @@ type CourseParticipation = {
   source?: string
 }
 
+type CourseParticipationUpdate = {
+  outcome: CourseParticipationOutcome
+  setting: CourseParticipationSetting
+  source: string
```
Gweaton, line 31: This could probably be `CourseParticipation['source']` too, in the spirit of the following two lines

Gweaton (approved):
Looks good. One tiny thing.

server/@types/models/CourseParticipation.ts
jsrobertson: Good spot

server/@types/models/CourseParticipation.ts
```diff
@@ -25,8 +25,22 @@ type CourseParticipation = {
   source?: string
 }
 
+type CourseParticipationUpdate = {
+  outcome: CourseParticipationOutcome
+  setting: CourseParticipationSetting
```
yndajas, line 30: Would this allow us to send an `undefined` setting, given `type` is a compulsory setting property? (C.f. all of the outcome properties are optional)

server/services/courseService.test.ts
```diff
@@ -133,6 +135,54 @@ describe('CourseService', () => {
...
+
+        await expect(() => service.getParticipation(token, courseParticipationId)).rejects.toHaveProperty(
+          'userMessage',
+          `Error fetching course participation with ID ${courseParticipationId}.`,
+        )
+
+        expect(courseClientBuilder).toHaveBeenCalledWith(token)
+        expect(courseClient.findParticipation).toHaveBeenCalledWith(courseParticipationId)
+      })
+    })
```
yndajas, line 183: Do we need these two tests? We don't have these error handling tests for other internal API service/client methods, only for the ones that hit external APIs (prisoner offender search and prison register)

server/services/courseService.ts
```diff
@@ -45,6 +48,29 @@ export default class CourseService {
...
+      const knownError = error as ResponseError
+
+      if (knownError.status === 404) {
+        return null
+      }
+
+      throw createError(knownError.status || 500, knownError, {
+        userMessage: `Error fetching course participation with ID ${courseParticipationId}.`,
+      })
+    }
```
yndajas, line 71: /do we need this error handling when we're working with our internal API (other internal API service/client methods don't do this)?

server/@types/models/CourseParticipation.ts
yndajas: Looks like `{}` is how we're meant to send `undefined` settings/outcomes: https://mojdt.slack.com/archives/C05045L6MEE/p1696001001594769

Gweaton: Oh yeah, we should update the `type` to be optional now. I'll add a card and open a PR for a couple of small fixes today

yndajas: A participation's `type` is now optional?

server/services/courseService.test.ts
jsrobertson: I think they're useful, why wouldn't we test them? Open to discussion though

server/services/courseService.ts
jsrobertson: Same as the tests really, I think it's worth adding some error handling.  I have actually updated this slightly in an upcoming PR/commit too as it's misusing `createError` and also think its' worth throwing an error if not found, instead of just `null`.

yndajas: Yeah maybe - I guess the reason we haven't done this so far is that we'd not expect our UI to show any data for the stuff we own if it doesn't exist in our API, so there shouldn't be 404s unless something's been deleted or the URL has been manually edited

Thinking about it, it probably does make more sense for participations since they're deletable, vs courses and referrals which are not (so you'd need to manually edit a URL or a form to try to get/post/put to a non-existent entity). We might want to add similar for `updateParticipation` in that case?

PR #246, opened 2023-10-04
(831) Handle errors in services

yndajas:
## Context

We're writing a lot of very similar error-handling tests because of resolving client errors in controllers

## Changes in this PR

- Simplify `organisationFromPrison` util method
- Handle errors in finding people/organisations in service
- Fix an error code

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/server/services/organisationService.ts
+++ b/server/services/organisationService.ts
@@ -18,7 +18,7 @@ export default class OrganisationService {
         return null
       }
 
-      return OrganisationUtils.organisationFromPrison(organisationId, prison)
+      return OrganisationUtils.organisationFromPrison(prison)
     } catch (error) {
       const knownError = error as ResponseError
 
--- a/server/services/organisationService.ts
+++ b/server/services/organisationService.ts
@@ -16,7 +16,7 @@ export default class OrganisationService {
       const prison = await prisonClient.find(organisationId)
 
       if (!prison.active) {
-        throw createError(500, `Organisation with ID ${organisationId} not active.`)
+        throw createError(404)
       }
 
       return OrganisationUtils.organisationFromPrison(prison)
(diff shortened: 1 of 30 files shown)


yndajas:
[Tidy up the way we're handling errors in our service](https://trello.com/c/7mD4fICN/831-tidy-up-the-way-were-handling-errors-in-our-service)

server/services/organisationService.ts
```diff
@@ -2,33 +2,66 @@ import createError from 'http-errors'
...
-  async getOrganisation(token: Express.User['token'], organisationId: string): Promise<Organisation | null> {
+  async getOrganisation(token: Express.User['token'], organisationId: Organisation['id']): Promise<Organisation> {
     const prisonClient = this.prisonClientBuilder(token)
 
     try {
       const prison = await prisonClient.find(organisationId)
 
       if (!prison.active) {
-        return null
+        throw createError(500, `Organisation with ID ${organisationId} not active.`)
```
Gweaton, line 19: I'm struggling to figure out what we should return here. What made you go for a 500? I think it's probably the most appropriate, I was tempted by a 422, or even a 404 but it's a bit of a hard one to pin down.

Gweaton (approved):
There's quite a lot to take in here, so hopefully I haven't missed anything, but this looks good to me. Just had one question about a choice of status code

server/services/organisationService.ts
Gweaton: Shall we go with 404 as a first step? I think you're right in saying the thing we're trying to find is active organisations (although that's not all that explicit from the URL) but I think, as you say, the message makes it clear.

yndajas: I wonder if the message should be `Active organisation with ID ... not found`?

Gweaton: Yeah that sounds good to me

PR #248, opened 2023-10-05
(853b) Add delete request support

yndajas:
## Context

#247 added the page for deleting course participations, but not the destroy action. Our codebase doesn't currently support DELETE requests

## Changes in this PR

- Add support for DELETE requests
- Add `RestClient` tests
- Remove retries on POST and PUT requests, per the template and Approved Premises

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/server/data/restClient.ts
+++ b/server/data/restClient.ts
@@ -160,10 +160,6 @@ export default class RestClient {
         .send(data)
         .agent(this.agent)
         .use(restClientMetricsMiddleware)
-        .retry(2, (err, res) => {
-          if (err) logger.info(`Retry handler found API error with ${err.code} ${err.message}`)
-          return undefined // retry handler only for logging retries, not to influence retry logic
-        })
         .auth(this.token, { type: 'bearer' })
         .set(headers)
         .responseType(responseType)
--- a/server/data/restClient.ts
+++ b/server/data/restClient.ts
@@ -18,6 +18,8 @@ interface GetRequest {
   responseType?: string
 }
 
+interface DeleteRequest extends GetRequest {}
+
 interface PostRequest {
   data?: Record<string, unknown>
   headers?: Record<string, string>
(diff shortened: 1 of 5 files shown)


server/data/restClient.ts
```diff
@@ -45,6 +47,38 @@ export default class RestClient {
...
+    logger.info(`Delete using user credentials: calling ${this.name}: ${path} ${query}`)
+    try {
+      const result = await superagent
+        .delete(`${this.apiUrl()}${path}`)
+        .agent(this.agent)
+        .use(restClientMetricsMiddleware)
+        .retry(2, (err, res) => {
+          if (err) logger.info(`Retry handler found API error with ${err.code} ${err.message}`)
+          return undefined // retry handler only for logging retries, not to influence retry logic
+        })
```
ushkarev, line 66: bear in mind that this means it retries twice unconditionally so if you don't want that behaviour, you might want to mimic the `retry = false` thingy in `requestWithBody`.

yndajas: Yeah, I think for now we're happy to keep this for GET and DELETE. It was a useful addition to the template though, so we'll be sure to steal it (and your other refactors) if our needs change!

ushkarev: i tend to cherry-pick updates to the template back to our repos, but it can be a total pain if we've changed a lot. i think ideally there could have been a base npm package with this kind of stuff.

>happy to keep this for GET and DELETE

at present, the (possibly-surprising!) retry behaviour is in DELTE but not in GET. it's in both in the template 🤷

yndajas: You mean in our repo? We have retries on GET here too

Yeah I think the optional approach is good! A straight cherry pick wouldn't quite here because we have hard mode on (`strict: true`), while the template doesn't 😅

yndajas: Yeah, we added it early on in the project. It was tricky getting everything to work, but not nearly as bad as adding it in much further down the line!

yndajas:
[Add page for deleting previous programme (course participation) (UI) (M)](https://trello.com/c/KwS7Yjeb/853-add-page-for-deleting-previous-programme-course-participation-ui-m)

PR #249, opened 2023-10-05
(888) Display not known against empty Course Participation values

Gweaton:
## Context

https://trello.com/c/0pCdIJDt/884-display-not-known-on-programme-history-when-the-user-hasnt-answered-a-question

Previously, we wouldn't show a row against a field that a user hadn't filled in when creating a CourseParticipation. We now would like to highlight this as "not known" to prompt users to enter any information they do know, rather than ignoring the field. We don't want to make the fields required, as this would prevent people who genuinely don't have this information from being able to add anything.

## Changes in this PR

- Displays "Not known" for rows where the CourseParticipation field is blank.
- Rejigs `CourseParticipationUtils` tests slightly to hopefully be a bit clearer now that this is the case.

### Before

Screenshot 2023-10-05 at 16 20 06

### After

image

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/server/utils/courseParticipationUtils.test.ts
+++ b/server/utils/courseParticipationUtils.test.ts
@@ -168,18 +168,6 @@ describe('CourseParticipationUtils', () => {
     })
 
     describe('when rows are missing required data', () => {
-      it.each([['Source of information', 'source', undefined]])(
-        'omits the %s row when %s is %s',
-        (keyText: GovukFrontendSummaryListRowKey['text'], field: string, value) => {
-          const withoutField = { ...courseParticipationWithName, [field]: value }
-
-          const { rows } = CourseParticipationUtils.summaryListOptions(withoutField, referralId)
-          const fieldRow = getRow(rows, keyText)
-
-          expect(fieldRow).toBeUndefined()
-        },
-      )
-
       describe('setting', () => {
         describe('when the setting has no location', () => {
           const withoutSettingLocation: CourseParticipationWithName = {
@@ -302,6 +290,20 @@ describe('CourseParticipationUtils', () => {
           it('displays "Not known"', () => {
             const { rows } = CourseParticipationUtils.summaryListOptions(withoutOutcomeAdditionalDetail, referralId)
             expect(getRowValueText(rows, 'Additional detail')).toEqual('Not known')
+          })
+        })
+      })
+
+      describe('source of information', () => {
+        describe('when there is no source of information set', () => {
+          const withoutSource: CourseParticipationWithName = {
+            ...courseParticipationWithName,
+            source: undefined,
+          }
+
+          it('displays "Not known"', () => {
+            const { rows } = CourseParticipationUtils.summaryListOptions(withoutSource, referralId)
+            expect(getRowValueText(rows, 'Source of information')).toEqual('Not known')
           })
         })
       })
--- a/server/utils/courseParticipationUtils.test.ts
+++ b/server/utils/courseParticipationUtils.test.ts
@@ -8,18 +8,11 @@ import { courseParticipationFactory } from '../testutils/factories'
 import type { CourseParticipationOutcome, CourseParticipationWithName } from '@accredited-programmes/models'
 import type { GovukFrontendSummaryListRow, GovukFrontendSummaryListRowKey } from '@govuk-frontend'
 
-const getRow = (
-  rows: Array<GovukFrontendSummaryListRow>,
-  keyText: GovukFrontendSummaryListRowKey['text'],
-): GovukFrontendSummaryListRow | undefined => {
-  return rows.find(row => (row.key as GovukFrontendSummaryListRowKey).text === keyText)
-}
-
 const getRowValueText = (
   rows: Array<GovukFrontendSummaryListRow>,
   keyText: GovukFrontendSummaryListRowKey['text'],
 ): string => {
-  return getRow(rows, keyText)!.value!.text!
+  return rows.find(row => (row.key as GovukFrontendSummaryListRowKey).text === keyText)!.value!.text!
 }
 
 describe('CourseParticipationUtils', () => {
--- a/server/utils/courseParticipationUtils.test.ts
+++ b/server/utils/courseParticipationUtils.test.ts
@@ -168,17 +168,17 @@ describe('CourseParticipationUtils', () => {
     })
 
     describe('when rows are missing required data', () => {
-      it.each([
-        ['Additional detail', 'outcome', { detail: undefined }],
-        ['Source of information', 'source', undefined],
-      ])('omits the %s row when %s is %s', (keyText: GovukFrontendSummaryListRowKey['text'], field: string, value) => {
-        const withoutField = { ...courseParticipationWithName, [field]: value }
+      it.each([['Source of information', 'source', undefined]])(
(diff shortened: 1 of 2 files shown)


server/utils/courseParticipationUtils.test.ts
```diff
@@ -176,28 +182,33 @@ describe('CourseParticipationUtils', () => {
...
-        const withoutSettingType = {
-          ...courseParticipationWithName,
-          setting: { location: 'Stockport', type: undefined },
-        }
-
-        const { rows } = CourseParticipationUtils.summaryListOptions(withoutSettingType, referralId)
-        const settingRow = getRow(rows, 'Setting')
-
-        expect(settingRow).toEqual({ key: { text: 'Setting' }, value: { text: 'Stockport' } })
-      })
```
yndajas: I guess this condition will be impossible when the API change to make `type` required again is merged, but until then and updates to the UI do we need to keep it? (setting location provided but setting type not - the util code supports this)

server/utils/courseParticipationUtils.test.ts
```diff
@@ -138,7 +138,7 @@ describe('CourseParticipationUtils', () => {
           },
           {
             key: { text: 'Outcome' },
-            value: { text: 'Complete - completed in 2019' },
+            value: { text: 'Complete, year completed 2019' },
```
yndajas: Do we really want it to look like that? Sounds/looks strange to me 😅

server/utils/courseParticipationUtils.test.ts
```diff
@@ -211,28 +209,70 @@ describe('CourseParticipationUtils', () => {
...
 
-        expect(outcomeRow).toEqual({ key: { text: 'Outcome' }, value: { text: 'Incomplete' } })
-      })
+          describe('and there is no yearStarted value', () => {
+            const withoutOutcomeYearStarted = {
+              ...courseParticipationWithName,
+              outcome: { status: 'incomplete' as CourseParticipationOutcome['status'], yearStarted: undefined },
+            }
+
+            it('displays the outcome on its own', () => {
```
yndajas, line 215: `status` or `outcome status`

server/utils/courseParticipationUtils.test.ts
```diff
@@ -211,28 +209,70 @@ describe('CourseParticipationUtils', () => {
...
-        }
+          describe('and there is no yearCompleted value', () => {
+            const withoutOutcomeYearCompleted = {
+              ...courseParticipationWithName,
+              outcome: { status: 'complete' as CourseParticipationOutcome['status'], yearCompleted: undefined },
+            }
 
-        const { rows } = CourseParticipationUtils.summaryListOptions(withoutOutcomeYearCompleted, referralId)
-        const outcomeRow = getRow(rows, 'Outcome')
+            it('displays the outcome and yearStarted value', () => {
```
yndajas, line 241: `'displays the outcome status on its own'` (or just `status` depending on previous comment)

server/utils/courseParticipationUtils.ts
```diff
@@ -91,27 +91,9 @@ export default class CourseParticipationUtils {
     })
 
     summaryListRows.push(CourseParticipationUtils.settingSummaryListRow(courseParticipationWithName))
+    summaryListRows.push(CourseParticipationUtils.outcomeSummaryListRow(courseParticipationWithName))
 
     if (courseParticipationWithName.outcome) {
```
yndajas: Gonna change anyway, but this outer condition can probably disappear in this commit (the second)

server/utils/courseParticipationUtils.ts
```diff
@@ -159,4 +141,28 @@ export default class CourseParticipationUtils {
       value: { text: valueTextItems.join(', ') || 'Not known' },
     }
   }
+
+  // eslint-disable-next-line @typescript-eslint/member-ordering
```
yndajas, line 119: Do you have a strong preference for having these methods listed in display order rather than the default alphabetical?

server/utils/courseParticipationUtils.ts
```diff
@@ -83,79 +83,95 @@ export default class CourseParticipationUtils {
...
+      valueTextItems.push(`${courseParticipationWithName.setting.location}`)
+    }
 
-      if (courseParticipationWithName.setting.location) {
-        valueTextItems.push(`${courseParticipationWithName.setting.location}`)
-      }
+    return {
+      key: { text: 'Setting' },
+      value: { text: valueTextItems.join(', ') || 'Not known' },
+    }
```
yndajas, line 116: I wonder if it's worth abstracting this into something like:

```ts
  private static summaryListRow(keyText: GovukFrontendSummaryListRowKey['text'], valueTextItems: Array<string>): GovukFrontendSummaryListRow {
    return {
      key: { text: keyText },
      value: { text: valueTextItems.join(', ') || 'Not known' },
    }
}
```

Then it can be reused in all of them including the `addedBy` one, and the programme name row could also use it

server/utils/courseParticipationUtils.ts
```diff
@@ -83,79 +83,95 @@ export default class CourseParticipationUtils {
...
+    ]
+  }
 
-    summaryListRows.push({
-      key: { text: 'Programme name' },
-      value: { text: courseParticipationWithName.name },
-    })
+  // eslint-disable-next-line @typescript-eslint/member-ordering
+  private static settingSummaryListRow(
+    courseParticipationWithName: CourseParticipationWithName,
```
yndajas, line 101: Is it worth making these method parameters a little narrower? E.g. this one could be `setting: CourseParticipationSetting`. It would mean a bit more indexing in the parent method, but less repetitive indexing within the methods

(I'm guessing we're passing around a reference rather than a pointer, so it's less of 'passing smaller things around' issue than a 'doing less indexing' and 'only giving methods what they need' issue)

yndajas (commented):
A few minor things and a couple of small refactor suggestions

server/utils/courseParticipationUtils.test.ts
Gweaton: You can't set the location but not the `type` in the UI, as you need to select the `type`'s conditional radio button to display the input for `location`. I suppose it's not clear from the implementation code though, we could wrap the `if (courseParticipationWithName.setting.location) ` inside the `if (courseParticipationWithName.setting.type) {` but this will change in future so we'll need to update it anyway.

I'll see what stage the API is at when I've made the other changes to this PR though.

server/utils/courseParticipationUtils.ts
Gweaton: I'd like them to be below `private static summaryListRows(`, if only we could have `private private` methods!

server/utils/courseParticipationUtils.test.ts
yndajas: True

server/utils/courseParticipationUtils.ts
yndajas: What if they were `summaryListRowsOutcome` etc instead of `outcomeSummaryListRow`, then we'd get it for free and also get the sorting check within the set of methods? It's more namespaced that way, which I think is the motivation for this sorting (?), but maybe less plain language

Gweaton:
Closing in favour of #256

PR #256, opened 2023-10-09
(888 - take 2) Display unanswered CourseParticipation fields as "Not known"

Gweaton:
## Context

This is a slight refactoring of #249 taking on @yndajas's helpful feedback. It felt easier to start from scratch and handle in one commit, rather than rebase the whole thing to make sense, as this is a more joined-up approach.

https://trello.com/c/0pCdIJDt/884-display-not-known-on-programme-history-when-the-user-hasnt-answered-a-question

Previously, we wouldn't show a row against a field that a user hadn't filled in when creating a CourseParticipation. We now would like to highlight this as "not known" to prompt users to enter any information they do know, rather than ignoring the field. We don't want to make the fields required, as this would prevent people who genuinely don't have this information from being able to add anything.

## Changes in this PR

- Displays "Not known" for rows where the CourseParticipation field is blank.
- Rejigs `CourseParticipationUtils` tests slightly to hopefully be a bit clearer now that this is the case.

### Before

image

### After

Screenshot 2023-10-09 at 16 05 59

Screenshot 2023-10-09 at 16 06 05

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/server/utils/courseParticipationUtils.test.ts
+++ b/server/utils/courseParticipationUtils.test.ts
@@ -5,18 +5,18 @@ import type { Request } from 'express'
 
 import CourseParticipationUtils, { type RequestWithDetailsBody } from './courseParticipationUtils'
 import { courseParticipationFactory } from '../testutils/factories'
+import type { CourseParticipationUpdate, CourseParticipationWithName } from '@accredited-programmes/models'
 import type {
-  CourseParticipationOutcome,
-  CourseParticipationSetting,
-  CourseParticipationUpdate,
-} from '@accredited-programmes/models'
-import type { GovukFrontendSummaryListRow, GovukFrontendSummaryListRowKey } from '@govuk-frontend'
+  GovukFrontendSummaryListRow,
+  GovukFrontendSummaryListRowKey,
+  GovukFrontendSummaryListRowValue,
+} from '@govuk-frontend'
 
-const getRow = (
+const getRowValueText = (
   rows: Array<GovukFrontendSummaryListRow>,
   keyText: GovukFrontendSummaryListRowKey['text'],
-): GovukFrontendSummaryListRow | undefined => {
-  return rows.find(row => (row.key as GovukFrontendSummaryListRowKey).text === keyText)
+): GovukFrontendSummaryListRowValue['text'] => {
+  return rows.find(row => (row.key as GovukFrontendSummaryListRowKey).text === keyText)!.value!.text!
 }
 
 describe('CourseParticipationUtils', () => {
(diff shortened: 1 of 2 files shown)


server/utils/courseParticipationUtils.ts
```diff
@@ -152,80 +153,64 @@ export default class CourseParticipationUtils {
...
-        valueTextItems.push(`${courseParticipationWithName.setting.location}`)
-      }
+  // eslint-disable-next-line @typescript-eslint/member-ordering
+  private static summaryListRow(
+    keyText: GovukFrontendSummaryListRowKey['text'],
+    valueTextItems: Array<string | undefined>,
+  ): GovukFrontendSummaryListRowWithValue {
+    return {
+      key: { text: keyText },
+      value: { text: valueTextItems.join(', ') || 'Not known' },
```
yndajas, line 176: If there's a `undefined` in here it will still add the `', '`, so you might want to filter out the `undefined` values before this

It doesn't look like any `undefined` values would make it here in an array of more than one item in the current state, but it might be good to tighten this in case things change

server/utils/courseParticipationUtils.ts
```diff
@@ -152,80 +153,64 @@ export default class CourseParticipationUtils {
...
-        valueText += StringUtils.properCase(courseParticipationWithName.outcome.status)
+    if (courseParticipationWithName.outcome?.yearStarted) {
+      valueTextItems.push(`Year started ${courseParticipationWithName.outcome.yearStarted}`)
+    }
 
-        if (courseParticipationWithName.outcome.yearStarted) {
-          valueText += ` - started ${courseParticipationWithName.outcome.yearStarted}`
-        }
+    if (courseParticipationWithName.outcome?.yearCompleted) {
+      valueTextItems.push(`Year complete ${courseParticipationWithName.outcome.yearCompleted}`)
```
yndajas, line 194: I know this is per the designs, but given the tense of `Year started` above, I think this should be `Year completed` for consistency

server/utils/courseParticipationUtils.test.ts
```diff
@@ -7,16 +7,16 @@ import CourseParticipationUtils, { type RequestWithDetailsBody } from './courseP
...
 } from '@accredited-programmes/models'
 import type { GovukFrontendSummaryListRow, GovukFrontendSummaryListRowKey } from '@govuk-frontend'
 
-const getRow = (
+const getRowValueText = (
   rows: Array<GovukFrontendSummaryListRow>,
   keyText: GovukFrontendSummaryListRowKey['text'],
-): GovukFrontendSummaryListRow | undefined => {
-  return rows.find(row => (row.key as GovukFrontendSummaryListRowKey).text === keyText)
+): string => {
```
yndajas, line 18: `GovukFrontendSummaryListRowValue['text']`?

server/utils/courseParticipationUtils.test.ts
```diff
@@ -354,68 +356,145 @@ describe('CourseParticipationUtils', () => {
...
-        const withoutSettingType = {
-          ...courseParticipationWithName,
-          setting: { location: 'Stockport', type: undefined },
-        }
+      describe('outcome', () => {
+        describe('when the outcome is incomplete', () => {
+          describe('and there is a yearStarted value', () => {
+            const withOutcomeYearStarted = {
+              ...courseParticipationWithName,
+              outcome: { status: 'incomplete' as CourseParticipationOutcome['status'], yearStarted: 2019 },
```
yndajas, line 394: If you wanted to avoid casting here and below you could type `withOutcomeYearStarted` as `CourseParticipationWithName` instead

server/utils/courseParticipationUtils.test.ts
```diff
@@ -354,68 +356,145 @@ describe('CourseParticipationUtils', () => {
...
+          })
+        })
       })
 
-      it('only shows the status in the Outcome row when yearStarted is undefined on an incomplete outcome', () => {
-        const withoutOutcomeYearStarted = {
-          ...courseParticipationWithName,
-          outcome: { status: 'incomplete' as CourseParticipationOutcome['status'], yearStarted: undefined },
-        }
+      describe('additional detail', () => {
```
yndajas, line 455: There's no "all information is provided" condition here (or for source below) like with setting and outcome. I guess it's included in the default test, but so are some of the setting and outcome combinations

server/utils/courseParticipationUtils.ts
Gweaton: Will leave this as-is after running it past Louise

server/utils/courseParticipationUtils.test.ts
yndajas: Makes sense

server/utils/courseParticipationUtils.ts
Gweaton: Good point, I've tightened this up, a bit of a hard one to test as we have no arrays that could possibly have more than one undefined element here (and it's a private method), so haven't updated tests for this.

PR #260, opened 2023-10-10
Update docker-compose.yml

bell-pa:
Add a second Postgresql database:
* prison-register-db for prison-register-api and 
* accredited-programmes-db for hmpps-accredited-programmes-api. 

Both databases listen on port 5432, but having different host names means that the services are able to select a database to connect to using its host name. Database ports are also exposed to the host environment (your laptop) at 5432 for accredited-programmes-api and 5433 for prison-register-api

There are no deployable (production) changes in the PR.


jsrobertson (approved):
Thanks @bell-pa !

PR #269, opened 2023-10-12
Tidy up types

yndajas:
## Changes in this PR

A couple of type-related tidy ups

I wonder if we could clarify what the `server/index.d.ts` and `users/index.d.ts` should be used for/why we need them alongside `ui/index.td.ts` and the `models` folder? Perhaps one for the upcoming ADRs

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/integration_tests/tsconfig.json
+++ b/integration_tests/tsconfig.json
@@ -10,7 +10,6 @@
     "paths": {
       "@accredited-programmes/integration-tests": ["./@types/index.d.ts"],
       "@accredited-programmes/models": ["../server/@types/models/index.d.ts"],
-      "@accredited-programmes/server": ["../server/@types/server/index.d.ts"],
       "@accredited-programmes/ui": ["../server/@types/ui/index.d.ts"],
       "@accredited-programmes/users": ["../server/@types/users/index.d.ts"],
       "@govuk-frontend": ["../server/@types/govukFrontend/index.d.ts"],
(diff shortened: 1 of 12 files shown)


Gweaton:
Re:
> I'm a little unclear why we have `server/index.d.ts` and `users/index.d.ts`
instead of using `ui/index.d.ts` for anything internal to logic within this
repo, and what those files should be used for. Perhaps we could resimplify which
might allow us to use a working aliased module in the Express type definitions
file

I agree, it's not that clear, and I think possibly undermines the need for index files. When adding the `users/index.d.ts` file, I did think about using `ui/index.d.ts`, but when looking at the types in that file they seem to be mostly related to manipulating data into a shape for presenting somewhere in the UI, so it didn't feel right to slot something relating to users in that file. `server/index.d.ts`, I'm not 100% sure on what else might be included here - given that it's to do with users' roles & permissions, maybe it could be moved into the `users/index.d.ts` file?

yndajas:
> maybe it could be moved into the `users/index.d.ts` file?

@Gweaton I've revised this PR to do this and fix another issue

integration_tests/tsconfig.json
```diff
@@ -12,7 +12,7 @@
       "@accredited-programmes/models": ["../server/@types/models/index.d.ts"],
       "@accredited-programmes/server": ["../server/@types/server/index.d.ts"],
       "@accredited-programmes/ui": ["../server/@types/ui/index.d.ts"],
-      "@accredited-programmes/users": ["./server/@types/users/index.d.ts"],
+      "@accredited-programmes/users": ["../server/@types/users/index.d.ts"],
```
Gweaton, line 14: argh I've done this so many times, I was sure I fixed it this time!

Gweaton (approved):
How would you feel about writing up an ADR to clarify our approach to using index files and aliases in the codebase? Would be very helpful for onboarding new people to our code style as we can't handle this with linters easily.

PR #279, opened 2023-10-17
(968b) Show users name instead of username on programme history entries

jsrobertson:
## Context

We want to display the name of the user who has added a programme history record.

## Changes in this PR
Added `getUserFromUsername` to `UserService` with error handling. This is then called in the referral and course participation controller to get the user details of each user.

Added a `userFactory` to help with testing.

Added `CourseParticipationPresenter` UI type which replaces `CourseParticipationWithName`.

### Before
image

### After
image

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/server/controllers/refer/courseParticipationsController.ts
+++ b/server/controllers/refer/courseParticipationsController.ts
@@ -1,15 +1,17 @@
 import type { Request, Response, TypedRequestHandler } from 'express'
 
 import { referPaths } from '../../paths'
-import type { CourseService, PersonService, ReferralService } from '../../services'
+import type { CourseService, PersonService, ReferralService, UserService } from '../../services'
 import { CourseParticipationUtils, CourseUtils, FormUtils, TypeUtils } from '../../utils'
-import type { CourseParticipation, CourseParticipationWithName, ReferralUpdate } from '@accredited-programmes/models'
+import type { CourseParticipation, ReferralUpdate } from '@accredited-programmes/models'
+import type { CourseParticipationPresenter } from '@accredited-programmes/ui'
 
 export default class CourseParticipationsController {
   constructor(
     private readonly courseService: CourseService,
     private readonly personService: PersonService,
     private readonly referralService: ReferralService,
+    private readonly userService: UserService,
   ) {}
 
   create(): TypedRequestHandler<Request, Response> {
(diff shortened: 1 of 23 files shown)


server/controllers/refer/courseParticipationsController.ts
```diff
@@ -249,10 +251,10 @@ export default class CourseParticipationsController {
     }
   }
 
-  private async courseParticipationsWithNames(
+  private async presentCourseParticipations(
```
jsrobertson, line 254: This is also in `referralController`.  Are we happy to have this in a util file and use the services in there?

server/controllers/refer/courseParticipationsController.ts
```diff
@@ -261,10 +263,12 @@ export default class CourseParticipationsController {
           const course = await this.courseService.getCourse(token, courseParticipation.courseId)
           name = course.name
         } else {
-          name = courseParticipation.otherCourseName as CourseParticipationWithName['name']
+          name = courseParticipation.otherCourseName as CourseParticipationPresenter['name']
         }
 
-        return { ...courseParticipation, name }
+        const addedByUser = await this.userService.getUserFromUsername(token, courseParticipation.addedBy)
```
jsrobertson, line 269: I also wonder if it might be worth doing things like this in the (course) service and getting the user details in there and returning the new `addedByName` property as part of `CourseParticipation`. Sorry if this is something that's been discussed already!

server/controllers/refer/courseParticipationsController.ts
yndajas: Ahh, the original method will be removed in the branch I'm working on, which we'd hoped would make this problem go away (hence the temporary duplication). The reason the original method was in the controllers and not utils is because it's calling services - it felt a bit weird to be doing that kind of async work in utils. But it looks like this introduces a new service call, so we'll have to work out some way to DRY it up?

server/services/userService.ts
```diff
@@ -22,6 +25,25 @@ export default class UserService {
     return { ...user, caseloads, displayName: StringUtils.convertToTitleCase(user.name) }
   }
 
+  async getUserFromUsername(token: Express.User['token'], username: User['username']): Promise<User> {
```
jsrobertson, line 28: Should probably be used by `getCurrentUserWithDetails` too, rather than calling the client directly in there.

server/controllers/refer/courseParticipationsController.ts
jsrobertson: Yeah that's my feeling too. I think like we've mentioned [here](https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/pull/279#discussion_r1362364359) might be the way to go.

jsrobertson:
Replaced with #286

PR #336, opened 2023-11-08
(1066b) Show release dates when viewing referral

yndajas:
## Changes in this PR

- Add release dates to sentence information page

### Before

image

### After

image

## Release checklist

### Pre-merge

### Post-merge


yndajas:
[Add sentence dates to view referral page (M)](https://trello.com/c/spzNM0Om/1066-add-sentence-dates-to-view-referral-page-m)

server/utils/personUtils.test.ts
```diff
@@ -90,6 +90,146 @@ describe('PersonUtils', () => {
...
+          homeDetentionCurfewEligibilityDate: undefined,
+          paroleEligibilityDate: undefined,
+          sentenceExpiryDate: undefined,
+          tariffDate: undefined,
+        }
+
+        expect(PersonUtils.releaseDatesSummaryListRows(person)).toEqual([
+          {
+            key: { text: 'Conditional release date' },
+            value: { text: 'There is no record for this date' },
```
Gweaton, line 210: I thought we were leaving dates off if we didn't have a record of them. I can see the prototype says otherwise. Just wanted to make sure that was the more recent decision?

integration_tests/pages/refer/show/sentenceInformation.ts
```diff
@@ -25,6 +25,17 @@ export default class SentenceInformationPage extends Page {
     )
   }
 
+  shouldContainReleaseDatesSummaryCard(): void {
+    cy.get('[data-testid="release-dates-summary-card"]').then(summaryCardElement => {
+      this.shouldContainSummaryCard(
+        'Release dates',
+        [],
+        PersonUtils.releaseDatesSummaryListRows(this.person),
```
Gweaton, line 33: I know we've used `utils` like this in a couple of places, which I may have missed at the time they were merged into `main`, but if the implementation/shape of the return value to `PersonUtils.releaseDatesSummaryListRows` changes, this test would still pass - we're essentially testing the implementation, using the implementation here, which doesn't feel very robust.

I'd prefer if we could write out the expected summary list rows here for this as an integration test as it'd catch this if it had an unexpected return value.

Gweaton (approved):
Looks good, just one thing I think we should solve re: robustness of integration tests.

integration_tests/pages/refer/show/sentenceInformation.ts
yndajas: Is this definitely the approach we want to take (I feel like we might have agreed to take the `server/utils` approach at some point, but not documented it!)? We use utils in a lot of places in the integration tests, as do [Approved Premises](https://github.com/search?q=repo:ministryofjustice/hmpps-approved-premises-ui+path:integration_tests+server/utils&type=code) and [Temporary Accommodation](https://github.com/search?q=repo:ministryofjustice/hmpps-temporary-accommodation-ui%20path:cypress_shared%20server/utils&type=code). The [Approved Premises summary list rows integration test method](https://github.com/search?q=repo:ministryofjustice/hmpps-approved-premises-ui%20shouldcontainsummarylistitems&type=code) is used in a similar way (passing in utils output), and Temporary Accommodation do something similar with [embedded records](https://github.com/search?q=repo:ministryofjustice/hmpps-temporary-accommodation-ui%20assertdefinition&type=code)

I can see why we'd want the integration tests not to be aware of implementation details but we're doing that all over the place, so I think this might be a bit of a substantial refactor - we might need to reimplement some of the utils code that's already unit tested elsewhere to make this dynamic, or set up multiple scenarios for the different earliest release dates so that we can have more hardcoded values (and similar in a few other contexts), which might be a bit overkill here? Maybe better to add this as a backlog ticket if it's important enough?

server/utils/personUtils.test.ts
yndajas: Yep we discussed that at the design/dev chat on Wednesday

PR #349, opened 2023-11-14
(pre-1107) Update sentence and offence details naming

yndajas:
## Context

Recent work has moved towards using the Prison API's conventions when naming things related to data we get from it

## Changes in this PR

In line with recent work, this updates:
- the type, service, path, variable names etc to follow the Prison API type
- the client method to follow the Prison API's equivalent method/endpoint name, except with `find` instead of `get`

## Release checklist

### Pre-merge

### Post-merge


server/services/personService.ts
```diff
@@ -81,6 +81,29 @@ export default class PersonService {
     }
   }
 
+  async getOffenderSentenceAndOffences(
+    token: Express.User['token'],
```
Gweaton, line 85: This should be take username, rather than user token and pass it to the generated system token - see other functions here for how to generate this

Gweaton: Just realised this isn't a change introduced in this PR, but it's something we do need to fix.

yndajas: #350 opened to deal with this

jsrobertson (approved):

yndajas:
[Show message if there are no sentence details (S)](https://trello.com/c/9sPNYxka/1107-show-message-if-there-are-no-sentence-details-s)

PR #354, opened 2023-11-15
Update view referral side navigation so it works correctly with the `/assess` path

jsrobertson:
## Context

When viewing a referral on the assess path, the links in the sidebar don't work correctly. They link back to refer and also don't show the active state.

## Changes in this PR
Update `ReferralUtils.viewReferralNavigationItems` to use assess paths if current path begins with `/assess`.

### Before
image

### After
image

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/server/utils/referralUtils.test.ts
+++ b/server/utils/referralUtils.test.ts
@@ -1,6 +1,6 @@
 import CourseUtils from './courseUtils'
 import ReferralUtils from './referralUtils'
-import { referPaths } from '../paths'
+import { assessPaths, referPaths } from '../paths'
 import {
   courseFactory,
   courseOfferingFactory,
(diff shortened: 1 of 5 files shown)


server/utils/referralUtils.ts
```diff
@@ -145,22 +145,25 @@ export default class ReferralUtils {
   static viewReferralNavigationItems(
     currentPath: Request['path'],
     referralId: Referral['id'],
+    isAssess?: boolean,
   ): Array<MojFrontendSideNavigationItem> {
+    const paths = isAssess ? assessPaths : referPaths
```
yndajas, line 150: Given we already have `currentPath` in here, could it just use `startsWith` like elsewhere instead of taking another argument? That might make it a bit less of a Refer as default arrangement

server/utils/referralUtils.test.ts
```diff
@@ -247,7 +247,7 @@ describe('ReferralUtils', () => {
   })
 
   describe('viewReferralNavigationItems', () => {
-    it('returns navigation items for the view referral pages and sets the requested page as active', () => {
+    it('returns navigation items for the view referral pages with the refer paths and sets the requested page as active', () => {
       const mockReferralId = 'mock-referral-id'
```
yndajas, line 251: Might be worth nesting both conditions in a `describe` block because Refer isn't really a default journey (although I think it's okay to use it as the default in the controller tests!)

The `const mockReferralId` can move into the outer scope here

yndajas (approved):
Couple of suggestions but in principle this looks good

PR #355, opened 2023-11-15
(803b) Use system token rather than user token for calls to Accredited Programmes API - Course Client

Gweaton:
## Context

https://trello.com/c/VxKOqk1u/803-use-system-token-to-communicate-with-api-m

We've had a new role added to secure our API endpoints (`ACCREDITED_PROGRAMMES_API`). This will require authenticating with client credentials, so we need to use our system token rather than the usual user token.

## Changes in this PR

This is the second half of #343.

Uses the system token rather than the user token for the Course Client methods, and updates services & controllers to generate and pass the correct token. 

I've split this PR up into lots of commits to hopefully keep it easier to review, but I'll squash them together once it's been approved to ensure we don't have any failing tests on a commit.

There may be a few clashes with #350 here, particularly with renaming, so we'll need to tidy some things up depending on which one is merged first.

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/server/controllers/shared/referralsController.test.ts
+++ b/server/controllers/shared/referralsController.test.ts
@@ -74,15 +74,7 @@ describe('ReferralsController', () => {
       const requestHandler = controller.additionalInformation()
       await requestHandler(request, response, next)
 
-      expect(referralService.getReferral).toHaveBeenCalledWith(username, referral.id)
-      expect(courseService.getCourseByOffering).toHaveBeenCalledWith(username, referral.offeringId)
-      expect(courseService.getOffering).toHaveBeenCalledWith(username, referral.offeringId)
-      expect(personService.getPerson).toHaveBeenCalledWith(
-        username,
-        person.prisonNumber,
-        response.locals.user.caseloads,
-      )
-      expect(organisationService.getOrganisation).toHaveBeenCalledWith(userToken, organisation.id)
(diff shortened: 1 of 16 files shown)


server/services/courseService.ts
```diff
@@ -17,27 +17,35 @@ import type { GovukFrontendSummaryListWithRowsWithKeysAndValues } from '@accredi
...
+    const hmppsAuthClient = this.hmppsAuthClientBuilder()
+    const systemToken = await hmppsAuthClient.getSystemClientToken(username)
+    const courseClient = this.courseClientBuilder(systemToken)
+
     return courseClient.destroyParticipation(courseParticipationId)
   }
 
   async getAndPresentParticipationsByPerson(
+    username: Express.User['username'],
     token: Express.User['token'],
```
yndajas, line 49: Can the token be renamed here?

server/controllers/shared/referralsController.test.ts
```diff
@@ -186,4 +155,12 @@ describe('ReferralsController', () => {
       })
     })
   })
+
+  function assertSharedDataServicesAreCalledWithExpectedArguments() {
```
yndajas, line 159: In this commit message, this isn't really "setup"

yndajas (approved):
Looks good - just a couple of small things that I can address and rebase on my branch before merging

PR #377, opened 2023-11-28
(1167b) WET pact

yndajas:
## Context

Keeping the API and UI in line with Pact tests is becoming a burden, and the DRYness of the tests on the UI side is making it hard to spot the difference

Related API PR: https://github.com/ministryofjustice/hmpps-accredited-programmes-api/pull/178

## Changes in this PR

- WET Pact tests
- Linting changes from #378

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/server/data/accreditedProgrammesApi/referralClient.test.ts
+++ b/server/data/accreditedProgrammesApi/referralClient.test.ts
@@ -1,3 +1,4 @@
+import { faker } from '@faker-js/faker/locale/en_GB'
 import { Matchers } from '@pact-foundation/pact'
 import { pactWith } from 'jest-pact'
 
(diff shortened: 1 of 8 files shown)


server/data/accreditedProgrammesApi/referralClient.test.ts
```diff
@@ -93,8 +89,8 @@ pactWith({ consumer: 'Accredited Programmes UI', provider: 'Accredited Programme
 
     beforeEach(() => {
       provider.addInteraction({
-        state: `Referral summaries exist for an organisation with the ID ${organisationId}`,
-        uponReceiving: `A request for the summaries for an organisation with the ID ${organisationId}`,
+        state: 'Referral summaries exist for organisation a026cc07-8e0b-40dd-9b66-1b3dacecc63d',
```
Gweaton, line 92: Organisation IDs are the same as Prison IDs in our service (so three characters), and the referral is fetched through the Offering so for this could be `BWN` to match our test data, rather than this UUID.

server/data/accreditedProgrammesApi/referralClient.test.ts
```diff
@@ -93,8 +89,8 @@ pactWith({ consumer: 'Accredited Programmes UI', provider: 'Accredited Programme
...
+        uponReceiving: "A request for organisation a026cc07-8e0b-40dd-9b66-1b3dacecc63d's referral summaries",
         willRespondWith: {
           body: Matchers.like(paginatedReferralSummaries),
```
Gweaton, line 95: Although `BWN` only has one referral in the test data and therefore one referral summary, I believe us using `Matchers.like` here will mean the test should pass as it's looking for a similarly shaped response but it doesn't have to be exact.

Gweaton (commented):
Looks good, thanks for sorting this all out, it's a lot easier to compare to the API data. Have gone through each test and the data and it looks good to me as far as I've seen. Just one ID to update

server/data/accreditedProgrammesApi/referralClient.test.ts
```diff
@@ -68,20 +65,19 @@ pactWith({ consumer: 'Accredited Programmes UI', provider: 'Accredited Programme
...
+      const result = await referralClient.find('0c46ed09-170b-4c0f-aee8-a24eeaeeddaa')
 
       expect(result).toEqual(referral)
     })
   })
 
   describe('findReferralSummaries', () => {
-    const organisationId = 'a026cc07-8e0b-40dd-9b66-1b3dacecc63d'
     const paginatedReferralSummaries: Paginated<ReferralSummary> = {
       content: referralSummaryFactory.buildList(2),
```
Gweaton, line 82: Is it worth popping in the `status: REFERRAL_SUBMITTED` here to make it a bit clearer that these are the referral summaries we're looking for?

Gweaton (approved):
One last thing, otherwise good to go!

server/data/accreditedProgrammesApi/referralClient.test.ts
yndajas: Technically that's only in the condition with query parameters, I think? But I could add that in that condition

Gweaton: I think it might make it clearer, but not too precious about it

PR #383, opened 2023-11-29
Miscellaneous pact fixes

Gweaton:
## Context

The Pact tests have been failing for a while and are incredibly brittle due to the way we set up state on the API before running them and then create/update/delete entities with state that persists between tests. We should definitely refactor this behaviour as soon as possible.

## Changes in this PR

While we wait for time to refactor, this PR updates the Pact contract tests to:

- Ensure we're never combining nullable and non nullable fields (e.g. `secondaryContactEmail` and `yearStarted`/`yearCompleted` in expected array responses, as Pact doesn't handle these well.
- Ensure we're not making assertions on created/updated/deleted entities like referrals and offerings in tests. This is due to the way state is handled in the API, as it's set up at the beginning of the test run. This has been incredibly painful to debug, and makes me very keen to refactor that behaviour so state isn't reused between tests on the API.

I've tested these against the API tests locally and they're all passing.

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/server/data/accreditedProgrammesApi/courseClient.test.ts
+++ b/server/data/accreditedProgrammesApi/courseClient.test.ts
@@ -231,7 +231,10 @@ pactWith({ consumer: 'Accredited Programmes UI', provider: 'Accredited Programme
   })
 
   describe('findOffering', () => {
-    const courseOffering = courseOfferingFactory.build({ id: '790a2dfe-7de5-4504-bb9c-83e6e53a6537' })
+    const courseOffering = courseOfferingFactory.build({
+      id: '790a2dfe-7de5-4504-bb9c-83e6e53a6537',
+      secondaryContactEmail: 'nobody2-bwn@digital.justice.gov.uk',
+    })
 
     beforeEach(() => {
       provider.addInteraction({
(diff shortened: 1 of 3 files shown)


server/data/accreditedProgrammesApi/courseClient.test.ts
```diff
@@ -327,7 +333,11 @@ pactWith({ consumer: 'Accredited Programmes UI', provider: 'Accredited Programme
   })
 
   describe('updateParticipation', () => {
-    const courseParticipation = courseParticipationFactory.build({ id: '0cff5da9-1e90-4ee2-a5cb-94dc49c4b004' })
+    const courseParticipation = courseParticipationFactory.build({
+      id: '0cff5da9-1e90-4ee2-a5cb-94dc49c4b004',
```
jsrobertson, line 337: Does this need to be `cc8eb19e-050a-4aa9-92e0-c654e5cfe281`?

Gweaton: Yep, good spot!

PR #399, opened 2023-12-11
(1137b) My referrals case list

jsrobertson:
## Context

We need to add a new **My referrals** page show referrals for the logged in use.

## Changes in this PR

- Add new ReferCaseListController
- This uses the existing dashboard endpoint for the time-being, to get something displayed and working. Then in theory it should be a case of switching to the new endpoint when it exists.
- Only adds the Open tab/page
- Linked to from from the primary navigation.

Suggest we move the `/refer/new` files in another directory if we still think that's needed.

## Screenshots of UI changes

localhost_3000_refer_referrals_case-list_open_page=1

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/server/controllers/refer/caseListController.ts
+++ b/server/controllers/refer/caseListController.ts
@@ -1,0 +1,52 @@
+import type { Request, Response, TypedRequestHandler } from 'express'
+import createHttpError from 'http-errors'
+
+import type { ReferralService } from '../../services'
+import { CaseListUtils, PaginationUtils, TypeUtils } from '../../utils'
+import type { ReferralStatus } from '@accredited-programmes/models'
+import type { ReferralStatusGroup } from '@accredited-programmes/ui'
+
+export default class ReferCaseListController {
+  constructor(private readonly referralService: ReferralService) {}
+
+  show(): TypedRequestHandler<Request, Response> {
+    return async (req: Request, res: Response) => {
+      TypeUtils.assertHasUser(req)
+
+      const referralStatusGroups: Record<ReferralStatusGroup, Array<ReferralStatus>> = {
+        closed: [],
+        draft: ['referral_started'],
+        open: ['assessment_started', 'awaiting_assessment', 'referral_submitted'],
+      }
+
+      const { page } = req.query as Record<string, string>
+      const { activeCaseLoadId, username } = res.locals.user
+      const { referralStatusGroup } = req.params
+
+      const referralStatuses = referralStatusGroups[referralStatusGroup as ReferralStatusGroup]
+
+      if (!referralStatuses) {
+        throw createHttpError(404, 'Not found')
+      }
+
+      const paginatedReferralSummaries = await this.referralService.getReferralSummaries(username, activeCaseLoadId, {
+        page: page ? (Number(page) - 1).toString() : undefined,
+        status: CaseListUtils.uiToApiStatusQueryParam(referralStatuses.join(',')),
+      })
+
+      const pagination = PaginationUtils.pagination(
+        req.path,
+        [],
+        paginatedReferralSummaries.pageNumber,
+        paginatedReferralSummaries.totalPages,
+      )
+
+      return res.render('referrals/caseList/refer/show', {
+        pageHeading: 'My referrals',
+        pagination,
+        subNavigationItems: CaseListUtils.subNavigationItems(req.path),
+        tableRows: CaseListUtils.tableRows(paginatedReferralSummaries.content),
+      })
+    }
+  }
+}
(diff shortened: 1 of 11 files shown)


server/controllers/refer/caseListController.ts
```diff
@@ -0,0 +1,60 @@
...
+      res.redirect(referPaths.caseList.show({ referralState: 'open' }))
+    }
+  }
+
+  show(): TypedRequestHandler<Request, Response> {
+    return async (req: Request, res: Response) => {
+      TypeUtils.assertHasUser(req)
+
+      const referralStates: Record<string, Array<string>> = {
+        // closed: [],
```
jsrobertson, line 24: May remove these commented out lines before merge, same for `subNavigationItems`

server/controllers/refer/index.ts
```diff
@@ -0,0 +1,14 @@
+/* istanbul ignore file */
+
+import ReferCaseListController from './caseListController'
+import type { Services } from '../../services'
+
+const controllers = (services: Services) => {
```
yndajas, line 6: I think we should keep the `refer/new/index` controllers, but I wonder if this top-level `refer/index` should be responsible for the `referNewControllers` as well, e.g. by importing them and spreading them into the return object? So then we just have one `referControllers` to import and spread at the top-level index? It's a lot of hierarchy, but it would match the file structure

server/controllers/refer/caseListController.ts
yndajas: I don't have a strong feeling, given we know they're needed sooner or later

server/controllers/refer/caseListController.ts
```diff
@@ -0,0 +1,50 @@
...
+import { CaseListUtils, PaginationUtils, TypeUtils } from '../../utils'
+
+export default class ReferCaseListController {
+  constructor(private readonly referralService: ReferralService) {}
+
+  show(): TypedRequestHandler<Request, Response> {
+    return async (req: Request, res: Response) => {
+      TypeUtils.assertHasUser(req)
+
+      const referralStates: Record<string, Array<string>> = {
```
yndajas, line 23: - Should this be `referralStatuses`?
- `Array<ReferralStatus>`?
- I can't remember if you can use a non-primitive type as the first argument(?) in a `Record` type, but I wonder if it's worth adding a UI type for the three referral status groups?

server/paths/refer.ts
```diff
@@ -7,17 +7,24 @@ const newReferralPeoplePathBase = offeringReferralPathBase.path('people')
...
 const newReferralProgrammeHistoryProgrammePath = newReferralProgrammeHistoryShowPath.path('programme')
 const newReferralConfirmOasysPath = newReferralShowPath.path('confirm-oasys')
 const newReferralAdditionalInformationPath = newReferralShowPath.path('additional-information')
 
-const referralShowPathBase = referPathBase.path('referrals/:referralId')
+const referralShowPathBase = referralsPath.path(':referralId')
 
 export default {
+  caseList: {
+    show: caseListPath.path(':referralState'),
```
yndajas, line 27: Are we referring to the three referral status groupings as 'referral state'? I wonder if this might get confusing given how similar in meaning they are? Is `referralStatusGroup(ing)` any clearer (here and elsewhere)?

server/controllers/refer/caseListController.ts
```diff
@@ -0,0 +1,50 @@
...
+        // closed: [],
+        // draft: ['referral_started'],
+        open: ['assessment_started', 'awaiting_assessment', 'referral_submitted'],
+      }
+
+      const { page } = req.query as Record<string, string>
+      const { activeCaseLoadId, username } = res.locals.user
+      const { referralState } = req.params
+
+      const status = referralStates[referralState]
```
yndajas, line 33: Regardless of where we land with naming, this should probably be plural

server/views/referrals/caseList/refer/show.njk
```diff
@@ -0,0 +1,55 @@
+{% from "govuk/components/pagination/macro.njk" import govukPagination %}
+{% from "moj/components/sub-navigation/macro.njk" import mojSubNavigation %}
+{% from "govuk/components/table/macro.njk" import govukTable %}
```
yndajas, line 3: The alphabetical order has been disrupted!

yndajas (commented):
Looks good. Just a few small things

server/views/referrals/caseList/refer/show.njk
jsrobertson: 😱

server/paths/refer.ts
jsrobertson: I've gone for variations of referral status group.

yndajas (approved):
I was gonna say that type should be singular but you beat me to it!

PR #388, opened 2023-12-04
(1138) Paginate case list

yndajas:
## Context

Taking the foundations from #86. Another reference PR could be https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/1010 but there's a lot going on there

## Release checklist

[Release process documentation](../doc/how-to/perform-a-release.md)

As part of our continuous deployment strategy we must ensure that this work is
ready to be released once merged.

### Pre-merge

- [ ] There are changes required to the Accredited Programmes API for this change to work...
  - [ ] ... and they have been released to production already

### Post-merge

- [ ] This adds, extends or meaningfully modifies a feature...
  - [ ] ... and I have written or updated an end-to-end test for the happy path in the [Accredited Programmes E2E repo](https://github.com/ministryofjustice/hmpps-accredited-programmes-e2e)
- [ ] This makes new expectations of the API...
  - [ ] ... and I have notified the API developer(s) of changes to the contract tests (Pact), or the API is already compliant
- [ ] [Manually approve](../doc/how-to/perform-a-release.md#releasing-to-the-preprod-environment) release to preprod
- [ ] [Manually approve](../doc/how-to/perform-a-release.md#releasing-to-the-production-environment) release to prod

PR #400, opened 2023-12-11
(1137a) Update existing `CaseListController` to be `AssessCaseListController`

jsrobertson:
## Context

We need to add the case list for a users referrals.

## Changes in this PR
After much thinking, I think we need 2 case list controllers, 1 for assess and 1 for refer.

This PR is to update what we currently have, to be used by assess, in anticipation for the my referral related case lists.

See draft PR #399 for what will hopefully follow on from this PR.

Also happy to wait for #388 to be merged first then will handle the rebase.

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/server/controllers/index.ts
+++ b/server/controllers/index.ts
@@ -1,5 +1,6 @@
 /* istanbul ignore file */
 
+import assessControllers from './assess'
 import DashboardController from './dashboardController'
 import findControllers from './find'
 import referNewControllers from './refer/new'
@@ -11,6 +12,7 @@ export const controllers = (services: Services) => {
 
   return {
     dashboardController,
+    ...assessControllers(services),
     ...findControllers(services),
     ...referNewControllers(services),
     ...sharedControllers(services),
--- a/server/@types/ui/index.d.ts
+++ b/server/@types/ui/index.d.ts
@@ -84,7 +84,7 @@ type OrganisationWithOfferingEmailsPresenter = Organisation & {
 
 type ReferralSharedPageData = {
   courseOfferingSummaryListRows: Array<GovukFrontendSummaryListRowWithKeyAndValue>
-  navigationItems: Array<MojFrontendSideNavigationItem>
+  navigationItems: Array<MojFrontendNavigationItem>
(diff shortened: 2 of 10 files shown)


server/controllers/index.ts
```diff
@@ -11,6 +12,7 @@ export const controllers = (services: Services) => {
 
   return {
     dashboardController,
+    ...caseListControllers(services),
```
Gweaton, line 15: I'm torn between your approach here and having an `assessControllers` and a `referControllers` for the two journeys, the latter of which feels a bit more user friendly when thinking about the different journeys - would it be tricky to split them in this way? This is obviously further complicated by the existence of the `referNewControllers`

yndajas: Yeah I think I agree. Everything is nested by journey unless shared. So this would be `controllers/assess/CaseListController` and the Refer one would be `controllers/refer/caseListController`. I think we can just move `referNewControllers` up a level and rename it `referControllers`?

server/paths/refer.ts
```diff
@@ -15,9 +15,14 @@ const newReferralProgrammeHistoryProgrammePath = newReferralProgrammeHistoryShow
 const newReferralConfirmOasysPath = newReferralShowPath.path('confirm-oasys')
 const newReferralAdditionalInformationPath = newReferralShowPath.path('additional-information')
 
+const caseListPath = referPathBase.path('case-list')
```
yndajas, line 18: Will this mean the path will be `/refer/case-list`, while Assess is `/assess/referrals/case-list`? I know we talked about the pathing elsewhere, but should this be consistent with Assessif we're going for that path for now?

server/controllers/index.ts
jsrobertson: Yeah agreed, makes sense 👍

server/paths/refer.ts
jsrobertson: Yeah will make them consistent.  Will move this commit to the other PR though.

yndajas (approved):

PR #403, opened 2023-12-12
chore: minor clean up the scripts

andrewcharnley:
A little tidy up of the scripts, things I had issues with while getting started.

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -11,5 +11,5 @@ script/utils/launch-docker
 
 echo "==> Installing application dependencies..."
 
-nodenv install --skip-existing
+npm_config_yes=true npx @nodenv/nodenv install --skip-existing
 npm install
--- a/script/utils/start-backing-services
+++ b/script/utils/start-backing-services
@@ -1,40 +1,31 @@
 #!/bin/sh
 
-echo "==> Starting the backing services in Docker..."
-
 for arg in "$@"
 do
   if [[ $arg == "--mock-api" ]]; then
     MOCK_API=true
+    DOCKER_PARAM="--scale=hmpps-accredited-programmes-api=0 --scale=hmpps-accredited-programmes-db=0"
   fi
 done
 
+echo "==> Starting the backing services in Docker..."
+docker compose up $DOCKER_PARAM -d
+
+echo "==> Resetting stubs..."
+npx ts-node --transpile-only ./wiremock/scripts/resetStubs.ts > /dev/null
+
 if [ "$MOCK_API" = true ]; then
-  docker compose up --scale=hmpps-accredited-programmes-api=0 --scale=postgresql=0 -d
-
   echo "==> Stubbing API endpoints..."
-
-  npx ts-node --transpile-only ./wiremock/scripts/resetStubs.ts > /dev/null
   npx ts-node --transpile-only ./wiremock/scripts/stubApis.ts > /dev/null
-  npx ts-node --transpile-only ./wiremock/scripts/stubCaseloads.ts > /dev/null
-  npx ts-node --transpile-only ./wiremock/scripts/stubOffenceCodes.ts > /dev/null
-  npx ts-node --transpile-only ./wiremock/scripts/stubOffenderBookings.ts > /dev/null
-  npx ts-node --transpile-only ./wiremock/scripts/stubPrisoners.ts > /dev/null
-  npx ts-node --transpile-only ./wiremock/scripts/stubOffenderSentenceAndOffences.ts > /dev/null
-
-else
-  docker compose up -d
-
-  npx ts-node --transpile-only ./wiremock/scripts/resetStubs.ts > /dev/null
-
-  echo "==> Stubbing Prison API caseload endpoints..."
-  npx ts-node --transpile-only ./wiremock/scripts/stubCaseloads.ts > /dev/null
-
-  echo "==> Stubbing Prisoner Search endpoints..."
-  npx ts-node --transpile-only ./wiremock/scripts/stubPrisoners.ts > /dev/null
-
-  echo "==> Stubbing Prison API endpoints..."
-  npx ts-node --transpile-only ./wiremock/scripts/stubOffenceCodes.ts > /dev/null
-  npx ts-node --transpile-only ./wiremock/scripts/stubOffenderBookings.ts > /dev/null
-  npx ts-node --transpile-only ./wiremock/scripts/stubOffenderSentenceAndOffences.ts > /dev/null
 fi
+
+echo "==> Stubbing Prison API caseload endpoints..."
+npx ts-node --transpile-only ./wiremock/scripts/stubCaseloads.ts > /dev/null
+
+echo "==> Stubbing Prisoner Search endpoints..."
+npx ts-node --transpile-only ./wiremock/scripts/stubPrisoners.ts > /dev/null
+
+echo "==> Stubbing Prison API endpoints..."
+npx ts-node --transpile-only ./wiremock/scripts/stubOffenceCodes.ts > /dev/null
+npx ts-node --transpile-only ./wiremock/scripts/stubOffenderBookings.ts > /dev/null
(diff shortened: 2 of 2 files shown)


script/utils/start-backing-services
```diff
@@ -1,40 +1,30 @@
...
-  docker compose up --scale=hmpps-accredited-programmes-api=0 --scale=postgresql=0 -d
+echo "==> Starting the backing services in Docker..."
+docker compose up $DOCKER_PARAM -d
 
-  echo "==> Stubbing API endpoints..."
+npx ts-node --transpile-only ./wiremock/scripts/resetStubs.ts > /dev/null
 
-  npx ts-node --transpile-only ./wiremock/scripts/resetStubs.ts > /dev/null
+if [ "$MOCK_API" = true ]; then
+  echo "==> Stubbing API endpoints..."
```
yndajas, line 17: Nice one. We can probably update this `echo` to `Stubbing Accredited Programme API endpoints...` for clarity

script/utils/start-backing-services
```diff
@@ -1,40 +1,30 @@
...
   fi
 done
 
-if [ "$MOCK_API" = true ]; then
-  docker compose up --scale=hmpps-accredited-programmes-api=0 --scale=postgresql=0 -d
+echo "==> Starting the backing services in Docker..."
+docker compose up $DOCKER_PARAM -d
 
-  echo "==> Stubbing API endpoints..."
+npx ts-node --transpile-only ./wiremock/scripts/resetStubs.ts > /dev/null
```
yndajas, line 14: Might be worth adding an echo to say `Resetting API stubs...` above this. In case anything fails, it'll help us pinpoint whether it's a Docker or Wiremock issue

package.json
```diff
@@ -7,21 +7,21 @@
   "scripts": {
     "build": "npm run compile-sass && tsc && npm run copy-views",
     "clean": "rm -rf dist build node_modules stylesheets",
-    "compile-sass": "sass --quiet-deps --no-source-map --load-path=node_modules/govuk-frontend --load-path=node_modules/@ministryofjustice/frontend --load-path=. assets/scss/application.scss:./assets/stylesheets/application.css assets/scss/application-ie8.scss:./assets/stylesheets/application-ie8.css --style compressed",
+    "compile-sass": "npx sass --quiet-deps --no-source-map --load-path=node_modules/govuk-frontend --load-path=node_modules/@ministryofjustice/frontend --load-path=. assets/scss/application.scss:./assets/stylesheets/application.css assets/scss/application-ie8.scss:./assets/stylesheets/application-ie8.css --style compressed",
```
yndajas, line 10: Were these scripts not working for you without `npx`? Did you run `script/setup` before using these?

yndajas (commented):
Nice tidy ups! That stubbing code was definitely in need of a DRY up

Might be good to break this into two commits, with a brief explanation of what's going on in each, e.g.:
- "Explicitly use `npx` to run `package.json` scripts" - with some explanation as to why this is needed in the commit body. Would be good to confirm that this is needed and/or beneficial
- "DRY up backing services script" - with a list of the two things that are DRYed up in the commit body, e.g. "Docker compose command line arguments"; "Wiremock stubbing". This could be two commits for an easier-to-read diff, but it's probably fine in one

[This ADR](https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/blob/main/doc/adr/0006-use-git-commits-to-explain-our-work.md) provides a little context on our approach to commits

package.json
andrewcharnley: That's correct, though it's on the MoJ macbook. The other reason is npm by default will use the global version if installed, npx reverses the behaviour.

script/utils/start-backing-services
andrewcharnley: Is this all it's stubbing, as the file is called stubApis.ts?

yndajas: Yep - it's probably also worth either (1) breaking down the `stubApis` script by entity, like the other scripts, or (at least) (2) renaming `stubApis` to `stubApi` or the mouthful `stubCoursesOfferingsParticipationsReferrals` (and updating the `console.log` near the end). Happy for that to be a separate piece of work if it's unclear!

package.json
andrewcharnley: script/setup wouldn't need npx, what's the error you see?

Gweaton: The error is on `script/test`:
```
> hmpps-accredited-programmes-ui@0.0.1 lint:njk
> npx djlint .

npm ERR! could not determine executable to run
```

But is the `npx` addition needed after running `script/setup`?

andrewcharnley: I skipped script/setup as it runs script/bootstrap, which requires nodenv globally. So far I'm running script/utils/start-backing-services to start docker compose + mocks, and npm run start:dev

Just checking script/test now...

yndajas: [djLint requires Python](https://www.npmjs.com/package/djlint/v/1.6.0#-how) so it might not work too well with npx?

> So far I'm running script/utils/start-backing-services to start docker compose + mocks, and npm run start:dev

We typically run `script/server` to start up the server (we're using the [Scripts To Rule Them All](https://github.com/github/scripts-to-rule-them-all) pattern), which bundles everything you need into one script, rather than running `npm run start:dev` directly

Running `script/setup` initially, `script/update` if you need to bring in any dependency updates (including to Docker images) then `script/server` to load up the server should generally do the trick

andrewcharnley: It failed for me where script/test calls script/update which calls script/bootstrap, which uses the nodenv, another that needs an npx before it.

yndajas: Ahh I see - are you using an alternative to `nodenv`?

yndajas: (nodenv is always installed globally - it's not an npm package)

andrewcharnley: Sorted, I'll leave the package.json alone for now, there's too many commands that I need to begin using first.

script/bootstrap
```diff
@@ -11,5 +11,5 @@ script/utils/launch-docker
 
 echo "==> Installing application dependencies..."
 
-nodenv install --skip-existing
-npm install
+npm_config_yes=true npx nodenv install --skip-existing
+npm i
```
yndajas, line 15: I'm still a little confused about this - are you able to document what's going on here and why this is needed in the commit message (same with the other changes in line with the ADR I linked in a separate comment)?

I didn't realise you could use `nodenv` via `npx`, but if we are going down that route (rather than installing via Homebrew), it looks like the [@nodenv/nodenv](https://www.npmjs.com/package/@nodenv/nodenv) package is the one, not [nodenv](https://www.npmjs.com/package/nodenv)?

Could we keep the more verbose `npm install` for readability?

script/utils/start-backing-services
andrewcharnley: Not touching it as part of this, it's down the rabbit hole.

script/bootstrap
andrewcharnley: Homebrew is mac specific and we should be environment agnostic.

Good spot on @nodenv/... Shame the author didn't have his original deprecated as I didn't get a warning.

It should be able to disappear though, we have docker for disconnecting the environment and can just make a override for app in docker-compose that runs nodemon inside the container.

yndajas: Could we resolve the `npm i` -> `npm install` issue and document in the commit message what's going on/why with the change on the previous line?

I'm also a little unclear on what the changes have to do with a container name, which is mentioned in the commit message - can you elaborate? As far as I can tell, the Docker change is just a DRY up

Gweaton (approved):

yndajas (commented):
Would be good to update the first line of the commit message to something like "Tidy up scripts" to make it easier to get a broad idea of what this is when searching the git history, but otherwise I think this is good to go

Gweaton:
@andrewcharnley Just spotted you'd merged main into this branch - we prefer to use rebase, so quickly un-automerged this and rebased (and updated the commit while I was at it) for future reference, sorry for treading on your toes, didn't want this to be a back and forth!

PR #405, opened 2023-12-13
(1099) Update refer journey input size boxes to match prototype

andrewcharnley:
Update refer journey input size boxes to match prototype

Requirements: https://trello.com/c/ritmihM7/1099-change-new-referral-input-sizes-xs

## Prototype

prison_number
programme-add
programme-detail-add

## Now

prison_number_new
programme-add-new
programme-detail-add-new

### DIFF
--- a/server/views/referrals/new/courseParticipations/course.njk
+++ b/server/views/referrals/new/courseParticipations/course.njk
@@ -24,9 +24,9 @@
     name: "otherCourseName",
     value: formValues.otherCourseName,
     type: "text",
-    classes: "govuk-!-width-one-third",
+    classes: "govuk-!-width-two-thirds",
     label: {
-      text: "Enter the programme"
+      text: "Enter the programme name"
     },
     errorMessage: errors.messages.otherCourseName
   }) }}
--- a/server/views/referrals/new/courseParticipations/details/show.njk
+++ b/server/views/referrals/new/courseParticipations/details/show.njk
@@ -22,7 +22,7 @@
 {{ govukInput({
   id: "communityLocation",
   name: "setting[communityLocation]",
-  classes: "govuk-!-width-one-third",
+  classes: "govuk-!-width-two-thirds",
   label: {
     text: "Enter the location (if known)"
   },
@@ -34,7 +34,7 @@
 {{ govukInput({
   id: "custodyLocation",
   name: "setting[custodyLocation]",
-  classes: "govuk-!-width-one-third",
+  classes: "govuk-!-width-two-thirds",
   label: {
     text: "Enter the prison (if known)"
   },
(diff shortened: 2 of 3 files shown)


server/views/referrals/new/courseParticipations/course.njk
```diff
@@ -24,9 +24,9 @@
     name: "otherCourseName",
     value: formValues.otherCourseName,
     type: "text",
-    classes: "govuk-!-width-one-third",
+    classes: "govuk-input--width-20",
```
yndajas, line 27: Looks like this is a bit wider on the prototype, using `govuk-!-width-two-thirds` - have you tried that here? Would probably make sense to be a bit wider

server/views/referrals/new/courseParticipations/course.njk
```diff
@@ -24,9 +24,9 @@
...
     label: {
-      text: "Enter the programme"
+      text: "Enter the programme name"
```
yndajas, line 29: Could we make this a separate commit or at least mention it in the commit message?

yndajas (commented):
Couple of small things here. Also, the conditionally-revealed setting and outcome inputs on the [Add programme details](https://accredited-programmes-prototype-main.apps.live.cloud-platform.service.justice.gov.uk/refer/v7/add-programme-details) page should be updated as part of this card

Would you be able to include the card number from Trello in the PR name to make it easier to cross-reference?

server/views/referrals/new/courseParticipations/course.njk
andrewcharnley: conditionally-revealed setting and outcome inputs  > already done, I checked them.

server/views/referrals/new/courseParticipations/course.njk
andrewcharnley: We could, it could also be two separate tickets... there's a yak that needs shaving! ;)

Gweaton: Let's add this to the commit message or get rid of it and tackle it in a separate card, as you say, so future readers inspecting the git history of this codebase can see when and why it changed

server/views/referrals/new/courseParticipations/course.njk
andrewcharnley: I'll double check as I recall the year already being 4 chars long.

yndajas (commented):
And here a first commit message line along the lines of "Update referral form presentation" would help it easier to get a broad idea of what the commit is about in the git history

Gweaton (approved):
Looks good.

As I think Ynda's mentioned on another PR, we tend not to use git flow in our commit messages, so in future we don't really need to add something like `style: trello card number` - we prefer to put trello context in the PR, so it doesn't clog up the initial line (helpful when running something like `git log --oneline` to see changes easily).

Here's some useful MOJ guidance for how to write commits https://technical-guidance.service.justice.gov.uk/documentation/guides/using-git.html#using-git which links to https://gds-way.digital.cabinet-office.gov.uk/standards/source-code/working-with-git.html#structure which might be helpful reading.

PR #409, opened 2023-12-14
(1138) Paginate case list

yndajas:
## Context

A review of a work-in-progress version of this can be found at #388

## Changes in this PR

- Add pagination utils
- Add pagination to Assess case list

## Screenshots of UI changes

image

image

image

## Release checklist

[Release process documentation](../doc/how-to/perform-a-release.md)

As part of our continuous deployment strategy we must ensure that this work is
ready to be released once merged.

### Pre-merge

- [ ] There are changes required to the Accredited Programmes API for this change to work...
  - [ ] ... and they have been released to production already

### Post-merge

- [ ] This adds, extends or meaningfully modifies a feature...
  - [ ] ... and I have written or updated an end-to-end test for the happy path in the [Accredited Programmes E2E repo](https://github.com/ministryofjustice/hmpps-accredited-programmes-e2e)
- [ ] This makes new expectations of the API...
  - [ ] ... and I have notified the API developer(s) of changes to the contract tests (Pact), or the API is already compliant
- [ ] [Manually approve](../doc/how-to/perform-a-release.md#releasing-to-the-preprod-environment) release to preprod
- [ ] [Manually approve](../doc/how-to/perform-a-release.md#releasing-to-the-production-environment) release to prod

PR #408, opened 2023-12-14
(pre-1138) Tidy up

yndajas:
## Changes in this PR

A few tidy ups made in the process of writing #409, including:
- updating Prison API types (this will hopefully stop being as much of a thing in random PRs after #348 is in
- mocking in unit tests
- a factory helper
- simplifying and tightening integration tests

## Release checklist

[Release process documentation](../doc/how-to/perform-a-release.md)

As part of our continuous deployment strategy we must ensure that this work is
ready to be released once merged.

### Pre-merge

- [ ] There are changes required to the Accredited Programmes API for this change to work...
  - [ ] ... and they have been released to production already

### Post-merge

- [ ] This adds, extends or meaningfully modifies a feature...
  - [ ] ... and I have written or updated an end-to-end test for the happy path in the [Accredited Programmes E2E repo](https://github.com/ministryofjustice/hmpps-accredited-programmes-e2e)
- [ ] This makes new expectations of the API...
  - [ ] ... and I have notified the API developer(s) of changes to the contract tests (Pact), or the API is already compliant
- [ ] [Manually approve](../doc/how-to/perform-a-release.md#releasing-to-the-preprod-environment) release to preprod
- [ ] [Manually approve](../doc/how-to/perform-a-release.md#releasing-to-the-production-environment) release to prod

PR #410, opened 2023-12-14
(1183) Filter out draft referrals from assess case list

jsrobertson:
## Context

On the assess case list view, we need to filter out any referrals with the REFERRAL_STARTED status.

## Changes in this PR
The dashboard endpoint now allows us to pass a list of statuses so we can filter by only the statuses we need.  The controller makes the request for all statuses except `referral_started` but will still use the selected filter value.

## Screenshots of UI changes

image

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/integration_tests/e2e/assess/caseList.cy.ts
+++ b/integration_tests/e2e/assess/caseList.cy.ts
@@ -13,11 +13,13 @@ context('Referral case lists', () => {
   const limeCourseReferralSummaries = FactoryHelpers.buildListWith(
     referralSummaryFactory,
     { courseName: limeCourse.name },
+    {},
     15,
   )
   const orangeCourseReferralSummaries = FactoryHelpers.buildListWith(
     referralSummaryFactory,
     { courseName: orangeCourse.name },
+    {},
     15,
   )
 
--- a/integration_tests/e2e/assess/caseList.cy.ts
+++ b/integration_tests/e2e/assess/caseList.cy.ts
@@ -10,16 +10,17 @@ context('Referral case lists', () => {
   const limeCourse = courseFactory.build({ name: 'Lime Course' })
   const orangeCourse = courseFactory.build({ name: 'Blue Course' })
   const courses = [limeCourse, orangeCourse]
+  const availableStatuses = ['assessment_started', 'awaiting_assessment', 'referral_submitted']
   const limeCourseReferralSummaries = FactoryHelpers.buildListWith(
     referralSummaryFactory,
     { courseName: limeCourse.name },
-    {},
+    { transient: { availableStatuses } },
     15,
   )
   const orangeCourseReferralSummaries = FactoryHelpers.buildListWith(
     referralSummaryFactory,
     { courseName: orangeCourse.name },
-    {},
+    { transient: { availableStatuses } },
     15,
   )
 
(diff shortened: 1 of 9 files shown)


integration_tests/e2e/assess/caseList.cy.ts
```diff
@@ -7,7 +7,6 @@ import type { Course, ReferralSummary } from '@accredited-programmes/models'
 
 const createReferralSummaries = (courseName: Course['name']): Array<ReferralSummary> => [
   referralSummaryFactory.build({ courseName, prisonNumber: 'ABC123', status: 'assessment_started' }),
-  referralSummaryFactory.build({ courseName, prisonNumber: 'ABC789', status: 'referral_started' }),
```
yndajas: Oops, after rebasing this with `main` (specifically [this change](https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/pull/408/commits/dedfb1aa64407ff114aec9a0fb077c925fa79d88#diff-b5ec79a3153a624b7738a4e5404fe792563df5716ef55a9c4769ddaeb4aec11cR12-R21)), this might need updating. I think we could:
- add a trait to the referral summary factory, which picks a random status other than `referral_started` (this feels like `submitted`, but in the referral factory `submitted` currently means one specific status 🤔)
- extend `buildListWith` to take an optional trait
- pass in the trait to `builtListWith` here

What do you think?

yndajas (commented):
Looks good. Something I did has probably made more work for you here - sorry! - but I think we should be able to continue down the same path and extend a helper function* to avoid having to write out lots of individual factory calls to build an array

\* that helper function is reused in #409, but in a different context

integration_tests/e2e/assess/caseList.cy.ts
yndajas: The approach I think you were taking just now (🦅) with the random status might also work, although invoking that function for each array item with `buildListWith` might be more of a challenge

jsrobertson: Opted for using `transientParams` in the end.  https://github.com/thoughtbot/fishery?tab=readme-ov-file#params-that-dont-map-to-the-result-object-transient-params

server/testutils/factories/referral.ts
```diff
@@ -35,13 +35,10 @@ class ReferralFactory extends Factory<Referral> {
...
 }
 
-export const randomStatus = () =>
-  faker.helpers.arrayElement([
-    'awaiting_assessment',
-    'assessment_started',
-    'referral_started',
-    'referral_submitted',
-  ]) as ReferralStatus
+export const randomStatus = (availableStatuses?: Array<string>) =>
```
yndajas, line 38: `Array<ReferralStatus>` here too

server/testutils/factories/referralSummary.ts
```diff
@@ -7,8 +7,13 @@ import { randomStatus } from './referral'
 import { StringUtils } from '../../utils'
 import type { ReferralSummary } from '@accredited-programmes/models'
 
-export default Factory.define<ReferralSummary>(({ params }) => {
-  const status = params.status || randomStatus()
+interface ReferralSummaryTransientParams {
+  availableStatuses: Array<string>
```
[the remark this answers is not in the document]
jsrobertson: I tend to go with Interface for object types (when there's no precedent in the same file) and then Type for anything else.  Not too precious either way though.

PR #415, opened 2023-12-19
(1198) Handle missing user loading

andrewcharnley:
When a user is not found a message is returned. The behaviour for two controllers was to throw if the user were missing.

Note the original function and its behaviour still exists as it may be used by other work in progress, and that behaviour may be preferred. If this is not the case a separate task could be actioned to remove it.

### Before

Screenshot 2023-12-19 at 11 53 08

### After

Screenshot 2023-12-19 at 11 50 38

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/server/services/userService.ts
+++ b/server/services/userService.ts
@@ -27,6 +27,15 @@ export default class UserService {
     return { ...user, caseloads, displayName: StringUtils.convertToTitleCase(user.name) }
   }
 
+  /*
+   * A helper which presents a consistent language string across controllers
+   * And which factors in the user may no longer be an entity
+   */
+  async getFullNameFromUsername(userToken: Express.User['token'], username: User['username']): Promise<string> {
+    const user = await this.checkUserExistsFromUsername(userToken, username)
+    return user ? StringUtils.convertToTitleCase(user.name) : `User '${username}' not found`
+  }
+
   async getUserFromUsername(userToken: Express.User['token'], username: User['username']): Promise<User> {
     const hmppsManageUsersClient = this.hmppsManageUsersClientBuilder(userToken)
 
@@ -39,11 +48,36 @@ export default class UserService {
         throw createError(knownError.status, `User with username ${username} not found.`)
       }
 
-      const errorMessage =
-        knownError.message === 'Internal Server Error' ? `Error fetching user ${username}.` : knownError.message
-
-      throw createError(knownError.status || 500, errorMessage)
+      throw this.createUserLoadNon404Error(username, knownError)
     }
+  }
+
+  /*
+   * Rather than return a boolean this function returns a thruthy by way of the User object
+   * This likely saves on a subsequent call afterwards to fetch these details.
+   */
+  private async checkUserExistsFromUsername(
+    userToken: Express.User['token'],
+    username: User['username'],
+  ): Promise<User | null> {
+    const hmppsManageUsersClient = this.hmppsManageUsersClientBuilder(userToken)
+
+    try {
+      const result = await hmppsManageUsersClient.getUserFromUsername(username)
+      return result
+    } catch (error) {
+      const knownError = error as ResponseError
+      if (knownError.status === 404) {
+        return null
+      }
+      throw this.createUserLoadNon404Error(username, knownError)
+    }
+  }
+
+  private createUserLoadNon404Error(username: string, knownError: ResponseError) {
+    const errorMessage =
+      knownError.message === 'Internal Server Error' ? `Error fetching user ${username}.` : knownError.message
+    return createError(knownError.status || 500, errorMessage)
   }
 
   private async getCaseloads(systemToken: SystemToken): Promise<Array<Caseload>> {
--- a/server/controllers/shared/referralsController.test.ts
+++ b/server/controllers/shared/referralsController.test.ts
@@ -82,7 +82,7 @@ describe('ReferralsController', () => {
     organisationService.getOrganisation.mockResolvedValue(organisation)
     personService.getPerson.mockResolvedValue(person)
     referralService.getReferral.mockResolvedValue(referral)
-    userService.getUserFromUsername.mockResolvedValue(referringUser)
+    userService.getFullNameFromUsername.mockResolvedValue(referringUser.name)
 
     controller = new SubmittedReferralsController(
       courseService,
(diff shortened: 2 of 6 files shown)


server/services/userService.ts
```diff
@@ -9,12 +9,37 @@ import type { SystemToken } from '@hmpps-auth'
...
+}
+
 export default class UserService {
   constructor(
     private readonly hmppsManageUsersClientBuilder: RestClientBuilder<HmppsManageUsersClient>,
     private readonly prisonApiClientBuilder: RestClientBuilder<PrisonApiClient>,
   ) {}
 
+  /*
+   * Rather than return a boolean this function returns a thruthy by way of the User object
```
Gweaton, line 25: Looks like you've got a typo in `truthy`, unless this is a word I didn't know!

Gweaton (approved):
Looks good - thanks for the descriptive commit message. Just one typo if you could amend that commit, then all good to merge.

Gweaton:
Ah looks like there's failing unit test too, actually

server/services/userService.ts
yndajas: Do we need the comment? I wonder if we could DRY up the overlap between this method and `getUserFromUsername`, if the difference is mostly just what to do in the case of a 404, depending on urgency here?

Gweaton: I agree, we've not used comments much to this effect in this codebase, and it looks a bit inconsistent. Let's get rid of them here.

server/services/userService.ts
```diff
@@ -9,12 +9,37 @@ import type { SystemToken } from '@hmpps-auth'
 import type { User } from '@manage-users-api'
 import type { Caseload } from '@prison-api'
 
+const throwUserLoadNon404Error = (username: string, knownError: ResponseError) => {
+  const errorMessage =
+    knownError.message === 'Internal Server Error' ? `Error fetching user ${username}.` : knownError.message
+  throw createError(knownError.status || 500, errorMessage)
+}
```
yndajas, line 16: This can probably be a private method on the class

server/services/userService.ts
andrewcharnley: > Looks like you've got a typo in `truthy`, unless this is a word I didn't know!

It's programming jargon for anything that resolves by the compiler to be true.

server/services/userService.ts
```diff
@@ -27,6 +52,15 @@ export default class UserService {
     return { ...user, caseloads, displayName: StringUtils.convertToTitleCase(user.name) }
   }
 
+  /*
+   * A helper which presents a consistent language string across controllers
+   * And which factors in the user may no longer be an entity
+   */
+  async getFullNameFromUsername(userToken: Express.User['token'], username: User['username']): Promise<string> {
+    const user = await this.checkUserExistsFromUsername(userToken, username)
```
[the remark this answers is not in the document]
Gweaton: I don't think there's a particular rush to get this in - I agree that it would be good to DRY this up to reduce the chance of people getting confused over which method to use.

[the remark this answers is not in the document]
andrewcharnley: > So I think we could probably DRY this up by adding a boolean argument to `getUserFromUsername` which determines whether the method errors on 404
> 
> ```ts
>   async getUserFromUsername(userToken: Express.User['token'], username: User['username'], errorOn404?: boolean = false): Promise<User> {
>       [...]
> 
>       if (knownError.status === 404) {
>         if (errorOn404) {
>           return null
>         } 
> 
>         throw createError(knownError.status, `User with username ${username} not found.`)
>       }
> 
>       [...]
>   }
> ```
> 
> This would mean we wouldn't need the `checkUserExistsFromUsername` method, and we can still use `getUserFromUsername` as before, but with the boolean set to `true` in `getFullNameFromUsername`. Then we can hopefully just add a condition to the `getUserFromUsername` tests to cover this
> 
> What do you think @andrewcharnley?

I consider it to be an anti-pattern, in FP approach one function does one thing well.

I was 50/50 on whether to rewrite the original function to change the behaviour but decided against it for the reason that the original decision on propagating the error and not going any further had a reasonable chance at being a business led one. If not, it was T.D, and a second quick task can now be to kill it. I wouldn't say there's a right or wrong here as both approaches will end up in the same place.

Small comments like those above the functions are to differentiate the distinction, they are the 'hints and tips' which make developers lives easier.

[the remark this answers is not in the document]
andrewcharnley: The original function is indeed still used and is a good example where behaviour may differentiate.

https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/blob/302102f8b20370f36a7490fa2fbe008b683a7020/server/services/userService.ts#L48

server/services/userService.ts
```diff
@@ -9,12 +9,37 @@ import type { SystemToken } from '@hmpps-auth'
...
   constructor(
     private readonly hmppsManageUsersClientBuilder: RestClientBuilder<HmppsManageUsersClient>,
     private readonly prisonApiClientBuilder: RestClientBuilder<PrisonApiClient>,
   ) {}
 
+  /*
+   * Rather than return a boolean this function returns a thruthy by way of the User object
+   * This likely saves on a subsequent call afterwards to fetch these details.
+   */
+  async checkUserExistsFromUsername(
```
Gweaton, line 28: I'm not sure I understand what this is used for - are we making a subsequent call after this (re: the "saves on subsequent call afterwards")?

Would combining the two functions into one work?

```
 async getUserNameOrUnknown(userToken: Express.User['token'], username: User['username']): Promise<string> {
    const hmppsManageUsersClient = this.hmppsManageUsersClientBuilder(userToken)

    let user: User | null

    try {
      user = await hmppsManageUsersClient.getUserFromUsername(username)
    } catch (error) {
      const knownError = error as ResponseError
      user = knownError.status === 404 ? null : throwUserLoadNon404Error(username, knownError)
    }
    return user ? StringUtils.convertToTitleCase(user.name) : `User '${username}' not found`
  }
```

Gweaton: OK, I'd like us not to keep going back and forth on this. To get this over the line, would you make `checkUserExistsFromUsername` method private, so it's clear to devs it's not used outside this class. That'll be enough to get this PR in.

Then (separate piece of work, I'll write a ticket) tidy up `getUserFromUsername` and this method so we don't have these duplicate methods hanging around (No 2 in the blog post you linked is "delete duplicate methods" after all).

Gweaton (approved):

PR #420, opened 2023-12-21
(1198) Add Basic Refer API Stubs

andrewcharnley:
As part of 1198 I attempted to mock the behaviour of a missing referer on local dev. Half way through a link came up on cloud dev which could be used to verify the ticket change and I also realised the refer dashboards are in a state of flux. This stub work is enough to get the refer overview up and running as it currently stands, with the four states (one for one org, three for another).


wiremock/scripts/stubApis.ts
```diff
@@ -1,11 +1,88 @@
...
+      const myReferrals = referralSummaries.filter(({ courseName }) => courseName === myCourseData.name)
+
+      // organisation/:orgId/courses?courseName=...
+      return () =>
+        stubFor({
+          request: {
+            method: 'GET',
+            url: `${apiPaths.referrals.dashboard({ organisationId })}?courseName=${encodeURIComponent(
+              myCourseData.name,
+            )}&size=999`,
```
Gweaton, line 64: Have just tested this out locally and we'll need to add `&status=ASSESSMENT_STARTED%2CAWAITING_ASSESSMENT%2CREFERRAL_SUBMITTED&size=15` to this, now that the view defaults to only open Referrals.

wiremock/scripts/stubApis.ts
```diff
@@ -1,11 +1,88 @@
 /* eslint-disable no-console */
 import { apiPaths } from '../../server/paths'
 import { stubFor } from '../index'
+import organizations from '../linkage/organizations.json'
 import { courseOfferings, courseParticipations, courses, prisoners, referrals } from '../stubs'
+import referralSummaries from '../stubs/referralSummaries.json'
```
Gweaton, line 6: Currently getting 404ing pages when clicking on links in the table, is that to be expected? Looks like we're not stubbing referral views here with your new referral summaries - do we need to add matching referrals to `referrals.json` for this, or is this PR just to get a basic page loading?

wiremock/linkage/organizations.json
```diff
@@ -0,0 +1,9 @@
+[
+  {
+    "id": "MRI",
```
Gweaton, line 3: We've gone with UK spelling for `organisations` elsewhere in the project, could we do that for this file? I think it's fine to pop `organisations.json` in the main `wiremock` directory, rather than nesting it in a `linkage` file too.

wiremock/scripts/stubApis.ts
```diff
@@ -22,7 +99,7 @@ stubs.push(async () =>
 )
 
 courses.forEach(course => {
-  stubs.push(async () =>
+  stubs.push(() =>
```
Gweaton, line 102: Ah, don't know why we thought we needed these to be async... well spotted.

Gweaton (commented):
Thanks for this, just a few minor things

wiremock/scripts/stubApis.ts
andrewcharnley: I only went so far with it to make the summary page load, the next step would be this.

wiremock/linkage/organizations.json
andrewcharnley: So wiremock/, not wiremock/stubs/ ?

(Just to be clear, this file is not a stub, it's the linkage between some stubs which builds the referral dynamically).

PR #424, opened 2024-01-03
(1126/1261) Add fields to case list

yndajas:
## Context

Closes #420

## Changes in this PR

- Tidy up stubbing and add case list 
- Add additional referral summary fields/case list columns
- Filter case list columns based on page

There's quite a lot here, with some chunky commits, so it might be good revise the history and split this into multiple PRs, but I'm out of time for this week

Still to do in a separate PR or two:
- use `referPaths` in `Name / Prison number` when on a `Refer` route
- add Refer integration tests

### Before

image

### After

image

image

image

## Release checklist

### Pre-merge

### Post-merge


yndajas:
[Add fields from external APIs to Caselists UI (L)](https://trello.com/c/LtYl9wt8/1126-add-fields-from-external-apis-to-caselists-ui-l)

yndajas:
[Remove "Date referred" column from Assess case list (S)](https://trello.com/c/GuYDs740/1261-remove-date-referred-column-from-assess-case-list-s)

server/utils/referrals/caseListUtils.ts
```diff
@@ -103,96 +103,112 @@ export default class CaseListUtils {
...
+            })
+            break
+          case 'Release date type':
+            row.push({
+              attributes: {
+                'data-sort-value': summary.sentence?.nonDtoReleaseDateType,
+              },
+              text: summary.sentence?.nonDtoReleaseDateType
+                ? {
+                    ARD: 'Automatic Release Date',
```
Gweaton: Might be worth extracting this out into some constants-type file as we may want to reference this elsewhere in future. It could live in the `utils` directory like [this one](https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/main/server/utils/allAPTypes.ts)

integration_tests/pages/shared/caseList.ts
```diff
@@ -44,24 +52,90 @@ export default class CaseListPage extends Page {
...
-          cy.get('.govuk-table__cell:nth-of-type(5)').should('have.html', CaseListUtils.statusTagHtml(summary.status))
+          this.columnHeaders.forEach((column, columnIndex) => {
+            cy.get('.govuk-table__cell')
+              .eq(columnIndex)
+              .then(tableCellElement => {
+                switch (column) {
+                  case 'Conditional release date':
+                    cy.wrap(tableCellElement).should(
+                      'have.text',
+                      CaseListUtils.tableRowContent(summary, 'Conditional release date'),
```
Gweaton, line 72: As I've said in a few PR comments, I'm not 100% confident on reusing implementing functions like `tableRowContent` here in the tests. If the implementation of that function changes, we won't catch it in our integration tests, although we _will_ catch it in our unit tests. I don't think this needs to block this PR, but I wonder if when we add the Refer tests we should split this `shared` test out into two tests and handwrite the different columns etc?

Gweaton (approved):
Looks like a lot of hard work, thanks for sorting all this. Have left a few fairly minor comments.

Gweaton:
Have also tested this out with the UI pointing at dev APIs and it's looking great! Only thing is for some reason names coming back uppercase, might be a dev data thing though.

image

server/testutils/factories/referralSummary.ts
```diff
@@ -21,8 +26,32 @@ export default Factory.define<ReferralSummary, ReferralSummaryTransientParams>((
       audience => audience.value,
     ),
     courseName: `${StringUtils.convertToTitleCase(faker.color.human())} Course`,
+    earliestReleaseDate: indeterminateSentence ? tariffExpiryDate : paroleEligibilityDate || conditionalReleaseDate,
+    organisationId: faker.string.alpha({ casing: 'upper', length: 3 }),
+    prisonName: Object.prototype.hasOwnProperty.call(params, 'prisonName')
```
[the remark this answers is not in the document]
yndajas: This is so that we can explicitly pass in `undefined` as a value

server/utils/referrals/caseListUtils.test.ts
```diff
@@ -159,126 +159,259 @@ describe('CaseListUtils', () => {
...
-            text: 'General offence, Extremism offence',
-          },
-          {
-            attributes: { 'data-sort-value': 'referral_submitted' },
-            html: CaseListUtils.statusTagHtml('referral_submitted'),
-          },
-        ],
-      ])
+    describe('when passed the headings for Assess case lists', () => {
+      it('formats referral summary information in the appropriate format for passing to a GOV.UK table Nunjucks macro, based on the provided headings', () => {
```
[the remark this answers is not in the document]
yndajas: Yeah we could do that - do you think it's okay just to have a single test with a subset of all possible columns passed in then?

server/utils/referrals/caseListUtils.ts
yndajas: Yeah - perhaps best to wait the need emerged though, unless we already have one planned?

yndajas:
> Have also tested this out with the UI pointing at dev APIs and it's looking great! Only thing is for some reason names coming back uppercase, might be a dev data thing though.

You mean person names, right? I don't remember noticing this in dev data elsewhere - maybe we do some explicit casing elsewhere though

integration_tests/pages/shared/caseList.ts
yndajas: We could, but we might end up needing to rewrite quite a bit of code every time we use it. We did use `CaseListUtils.statusTagHtml(summary.status)` here before. Maybe one for later if we want to revise our approach across the board in the integration tests

yndajas: We'd also need to do that for every row, but I suppose we can just use a small number of rows in the integration tests when not testing pagination

server/utils/referrals/caseListUtils.test.ts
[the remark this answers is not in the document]
Gweaton: I think so, yeah

server/testutils/factories/referralSummary.ts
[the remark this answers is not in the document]
Gweaton: ah, in that case, might be worth adding this to the commit message and/or encapsulating this in a private method like getPrisonNameOrUndefinedFromParams (or something less horrible) to make that obvious

server/utils/referrals/caseListUtils.ts
Gweaton: sure, I'm not sure if there's anywhere else we're displaying these values

integration_tests/pages/shared/caseList.ts
Gweaton: Yeah, I think this is ok for now, but when we split out the two pages we should write it out to make our tests a bit more robust

server/testutils/factories/referralSummary.ts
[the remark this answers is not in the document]
yndajas: This is actually asking "does this property exist on the object?" (differentiating `undefined` and not defined), rather than "get me the value even if undefined" - the latter (in this ternary context) would result in the same as what you wrote above:

```ts
params.prisonName || `${faker.location.county() (HMP)
```

We could add a factory helper method along the lines of `objectHasProperty(object, property)` although it might just be rewording the built-in library code?

integration_tests/pages/shared/caseList.ts
yndajas: Might be worth a quick huddle if you have a minute later just to make sure I know what you're suggesting before I write the Refer test!

server/utils/referrals/caseListUtils.test.ts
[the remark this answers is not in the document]
yndajas: I've revised this a little so that it checks all fields in one test (since we want to check the rest of the object looks right, not just the bits generated by `tableRowContent`), then checks that filtering works in another with just three columns passed in

PR #429, opened 2024-01-04
(1139b) Draft referral progress

jsrobertson:
## Context

We need to display how many tasks have been completed for a draft referral in the draft referrals case list table.

## Changes in this PR

- Adds new method to `ReferralService` to calculate how many of the required fields on a referral have truthy values.
- Above method called in `ReferCaseListController.show` to return a `tasksCompleted` value that we can use in the table template.

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/server/services/referralService.test.ts
+++ b/server/services/referralService.test.ts
@@ -112,6 +112,35 @@ describe('ReferralService', () => {
     })
   })
 
+  describe('getNumberOfTasksCompleted', () => {
+    let referralService: ReferralService
+
+    beforeEach(() => {
+      referralService = createMock<ReferralService>(service)
+    })
+
+    it.each`
+      referralFormFields                                                                                          | expectedTasksCompleted
+      ${{ additionalInformation: '', hasReviewedProgrammeHistory: false, oasysConfirmed: false }}                 | ${1}
+      ${{ additionalInformation: 'Some information', hasReviewedProgrammeHistory: false, oasysConfirmed: false }} | ${2}
+      ${{ additionalInformation: 'Some information', hasReviewedProgrammeHistory: true, oasysConfirmed: false }}  | ${3}
+      ${{ additionalInformation: 'Some information', hasReviewedProgrammeHistory: true, oasysConfirmed: true }}   | ${4}
+    `(
+      'returns $expectedTasksCompleted when $referralFormFields',
+      async ({ referralFormFields, expectedTasksCompleted }) => {
+        const referral = referralFactory.started().build({
+          ...referralFormFields,
+        })
+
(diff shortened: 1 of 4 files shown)


server/services/referralService.test.ts
```diff
@@ -112,6 +112,32 @@ describe('ReferralService', () => {
...
 
+  describe('getNumberOfTasksCompleted', () => {
+    let referralService: ReferralService
+
+    beforeEach(() => {
+      referralService = createMock<ReferralService>(service)
+    })
+
+    it.each`
+      referralKeys                                                                                                | expected
```
Gweaton, line 123: This table test is really cool. I found it a bit tricky to read at first, I wonder if we could make it even easier to scan with variable names like `referralFormFields` and `expectedTasksCompleted` (something a bit more verbose)?

server/services/referralService.ts
```diff
@@ -40,6 +40,18 @@ export default class ReferralService {
...
+  async getNumberOfTasksCompleted(username: Express.User['username'], referralId: Referral['id']): Promise<number> {
+    const referralProgressIndicatorKeys = [
+      'hasReviewedProgrammeHistory',
+      'oasysConfirmed',
+      'additionalInformation',
+    ] as const
+
+    const referral = await this.getReferral(username, referralId)
+
+    return 1 + referralProgressIndicatorKeys.filter(item => referral[item]).length
```
Gweaton, line 52: Would checking for the presence of a `prisonNumber` on the Referral be a bit clearer here than using `1`? The commit message does explain this, but might be easier for future readers to understand.

Gweaton (commented):
I think this looks good - just have a couple of small comments

yndajas:
@Gweaton @jsrobertson the need to make an extra call to the API per referral is making me think this should really be provided by the API in the `ReferralSummary`. An extra API 15 calls per case list page seems substantial  - thoughts?

Gweaton:
> @Gweaton @jsrobertson the need to make an extra call to the API per referral is making me think this should really be provided by the API in the `ReferralSummary`. An extra up to 15 API calls per case list page seems substantial - thoughts?

Yeah I did wonder about this, it doesn't feel ideal when we have this smaller `referralSummary` object... We might need to rethink this.

jsrobertson:
> @Gweaton @jsrobertson the need to make an extra call to the API per referral is making me think this should really be provided by the API in the `ReferralSummary`. An extra up to 15 API calls per case list page seems substantial - thoughts?

Yeah, I don't disagree to be fair but would be interesting to know how big these lists are likely to get and if it really is a issue in the grand scheme of things.

Gweaton (approved):

PR #183, opened 2023-09-08
(650b) Complete referral page

jsrobertson:
## Context

This page will be redirected to from an upcoming submit referral endpoint.

## Changes in this PR

- Updated `Referral` type to include `status`
- Added `feedbackUrl` as nunjucks global to be used in multiple places
- Add complete page template
- Added complete method to referral controller
- Added `/complete` path for referral journey

### Before
N/A

### After
localhost_3000_referrals_c11fab18-dc8d-420c-9c82-d0edd373732d_complete

## Release checklist

[Release process documentation](../doc/how-to/perform-a-release.md)

As part of our continuous deployment strategy we must ensure that this work is
ready to be released once merged.

### Pre-merge

- [ ] There are changes required to the Accredited Programmes API for this change to work...
  - [ ] ... and they have been released to production already

### Post-merge

- [ ] This adds, extends or meaningfully modifies a feature...
  - [ ] ... and I have written or updated an end-to-end test for the happy path in the [Accredited Programmes E2E repo](https://github.com/ministryofjustice/hmpps-accredited-programmes-e2e)
- [ ] This makes new expectations of the API...
  - [ ] ... and I have notified the API developer(s) of changes to the contract tests (Pact), or the API is already compliant
- [ ] [Manually approve](../doc/how-to/perform-a-release.md#releasing-to-the-preprod-environment) release to preprod
- [ ] [Manually approve](../doc/how-to/perform-a-release.md#releasing-to-the-production-environment) release to prod

PR #445, opened 2024-01-11
Remove random nulls from factories

Gweaton:
## Context

Our Pact contracts are being generated every time we deploy a branch to dev. Because we allow for randomly `null` values in our Factory-generated objects, it makes it impossible for the API code to reliably provide these.

## Changes in this PR

Removes random `null` setting of fields in our factories. We originally added this functionality to make our generated objects more realistic in relation to what would come back from the various APIs, but we're already manually passing in these `null` values in tests that check for them.

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/server/testutils/factories/courseParticipation.ts
+++ b/server/testutils/factories/courseParticipation.ts
@@ -3,7 +3,6 @@ import { Factory } from 'fishery'
 
 import courseParticipationOutcomeFactory from './courseParticipationOutcome'
 import courseParticipationSettingFactory from './courseParticipationSetting'
-import FactoryHelpers from './factoryHelpers'
 import { StringUtils } from '../../utils'
 import type { CourseParticipation } from '@accredited-programmes/models'
 
@@ -24,8 +23,8 @@ export default CourseParticipationFactory.define(() => ({
   courseName: `${StringUtils.convertToTitleCase(faker.color.human())} Course`,
   createdAt: `${faker.date.between({ from: '2023-09-20T00:00:00.000Z', to: new Date() })}`,
   detail: faker.lorem.paragraph({ max: 5, min: 0 }),
-  outcome: FactoryHelpers.optionalArrayElement([courseParticipationOutcomeFactory.build()]),
+  outcome: courseParticipationOutcomeFactory.build(),
   prisonNumber: faker.string.alphanumeric({ length: 7 }),
-  setting: FactoryHelpers.optionalArrayElement([courseParticipationSettingFactory.build()]),
-  source: FactoryHelpers.optionalArrayElement(faker.word.words()),
+  setting: courseParticipationSettingFactory.build(),
+  source: faker.word.words(),
 }))
--- a/server/testutils/factories/courseParticipationOutcome.ts
+++ b/server/testutils/factories/courseParticipationOutcome.ts
@@ -1,7 +1,6 @@
 import { faker } from '@faker-js/faker/locale/en_GB'
 import { Factory } from 'fishery'
 
-import FactoryHelpers from './factoryHelpers'
 import type { CourseParticipationOutcome } from '@accredited-programmes/models'
 
 const randomValidYear = (): number => {
@@ -13,7 +12,7 @@ const outcomeTypes = {
   complete(): CourseParticipationOutcome {
     return {
       status: 'complete',
-      yearCompleted: FactoryHelpers.optionalArrayElement(randomValidYear()),
+      yearCompleted: randomValidYear(),
       yearStarted: undefined,
     }
   },
@@ -22,7 +21,7 @@ const outcomeTypes = {
     return {
       status: 'incomplete',
       yearCompleted: undefined,
-      yearStarted: FactoryHelpers.optionalArrayElement(randomValidYear()),
+      yearStarted: randomValidYear(),
     }
   },
 
--- a/server/testutils/factories/courseParticipationSetting.ts
+++ b/server/testutils/factories/courseParticipationSetting.ts
@@ -1,10 +1,9 @@
 import { faker } from '@faker-js/faker/locale/en_GB'
 import { Factory } from 'fishery'
 
-import FactoryHelpers from './factoryHelpers'
 import type { CourseParticipationSetting } from '@accredited-programmes/models'
 
 export default Factory.define<CourseParticipationSetting>(() => ({
-  location: FactoryHelpers.optionalArrayElement(faker.location.city()),
+  location: faker.location.city(),
   type: faker.helpers.arrayElement(['custody', 'community']),
 }))
--- a/server/testutils/factories/factoryHelpers.ts
+++ b/server/testutils/factories/factoryHelpers.ts
@@ -28,7 +28,7 @@ export default class FactoryHelpers {
     return faker.helpers.arrayElement(fullOptions.concat(options))
   }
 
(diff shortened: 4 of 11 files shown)


server/testutils/factories/courseParticipation.ts
```diff
@@ -3,7 +3,6 @@ import { Factory } from 'fishery'
```
yndajas, line 3: As discussed, might be good to clarify "nullable or string data" as `string | undefined` (as opposed to `string`)

yndajas (commented):
Are we sure about this? Are the Pact tests so much of a nightmare that it's not easy enough just to set values in the client tests where absolutely needed, without making our factories less realistic?

To me, it feels like going against the principle we informally settled on [here](https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/pull/183#discussion_r1319891520) and [here](https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/pull/193#discussion_r1324149829). Assumption of non-nullability on an optional field doesn't feel meaningfully different to assumption of a particular value from a set of options

If we do need this/this is a much easier approach than isolating the solution to the client tests, it would probably be good to explain in one of the commits why the API can't handle the potential of 'null' data on optional fields (motivating this shift in principle), given this goes against the UI type and API schema definitions

Gweaton:
> Are we sure about this? Are the Pact tests so much of a nightmare that it's not easy enough just to set values in the client tests where absolutely needed, without making our factories less realistic?
> 
> To me, it feels like going against the principle we informally settled on [here](https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/pull/183#discussion_r1319891520) and [here](https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/pull/193#discussion_r1324149829). Assumption of non-nullability on an optional field doesn't feel meaningfully different to assumption of a particular value from a set of options
> 
> If we do need this/this is a much easier approach than isolating the solution to the client tests, it would probably be good to explain in one of the commits why the API can't handle the potential of nullish data on optional fields (motivating this shift in principle), given this goes against the UI type and API schema definitions

The Pact tests have definitely become unmanageable, and are taking days to fix - this has happened 3 times over the past sprint whenever a UI branch is deployed.

Having this updated at factory level will make it much easier going forward for UI devs to not need to think too hard when adding new tests and objects to the client tests, as they shouldn't break things on the API side.

Despite those previous decisions, I'm not sure I see the benefit of our factories having so much of this nullability logic in them - the fact that no tests have needed updating with these changes highlights the fact that we're not really gaining benefit from these fields being nullable.

I'm not sure I agree that this is actively going against the UI type and API schema definitions - we're using these as default values (essentially a "happy path") where no arguments are provided. We can override them where we do care about the values being nullable, which we're already doing in unit tests.

server/testutils/factories/courseParticipation.ts
Gweaton: Makes sense - have updating the commit message to include this and the Pact info in your other comment - hope it's a bit clearer!

yndajas: Is it: for optional string fields, Pact either wants a string or nothing, not `null` or `undefined`? The problem being we're sometimes sending `null` or `undefined` when it wants a string or nothing?

yndajas:
> The Pact tests have definitely become unmanageable, and are taking days to fix - this has happened 3 times over the past sprint whenever a UI branch is deployed.
> 
> Having this updated at factory level will make it much easier going forward for UI devs to not need to think too hard when adding new tests and objects to the client tests, as they shouldn't break things on the API side.

If [this](https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/pull/445#discussion_r1448747880) is correct, I think I might now have a vague understanding of why this is a problem for Pact and/or the API side, but it might be good to be even more explicit in the commit message!
 
> Despite those previous decisions, I'm not sure I see the benefit of our factories having so much of this nullability logic in them - the fact that no tests have needed updating with these changes highlights the fact that we're not really gaining benefit from these fields being nullable.
> 
> I'm not sure I agree that this is actively going against the UI type and API schema definitions - we're using these as default values (essentially a "happy path") where no arguments are provided. We can override them where we do care about the values being nullable, which we're already doing in unit tests.

It does seem like we're explicitly setting fields as `undefined` in relevant test conditions, but that's because the factories are only optionally undefined: inversely we're presumably also setting fields as non-nullish where needed (outside the client tests), so if we made the factories default to `undefined` on optional fields, there's a chance everything would also still work (but I won't bet on it 😅)!

I'm still struggling to see the difference between assuming non-nullability versus assuming one non-nullish value out of a set: neither should cause errors for optional fields, so both seem like happy paths to me. I don't mind going down this route for now if it's definitely the quickest way to unblock things at this stage, but it might be worth adding a ticket to the backlog to add an ADR on our use of factories - we can assume non-nullability on optional fields because X (even if X is just "it causes fewer headaches with Pact, which we value more than optimal realism"), but we can't assume a particular non-nullish value out of a set of options because Y - given it feels like a little bit of a grey area?

server/testutils/factories/courseParticipation.ts
Gweaton: What happens is:

- UI client test generates an object with an undefined value (e.g. `CourseParticipation.setting`), then generates the JSON contract file expecting the API to send an object with the undefined value (or the field missing), e.g - second object here is missing the `setting` object:

```
{
  "status": 200,
  "headers": {
    "Content-Type": "application/json"
  },
  "body": [
    {
      "addedBy": "Devin Bins",
      "courseName": "Red Course",
      "createdAt": "Fri Nov 24 2023 03:46:19 GMT+0000 (Coordinated Universal Time)",
      "detail": "",
      "id": "0cff5da9-1e90-4ee2-a5cb-94dc49c4b004",
      "outcome": {
        "status": "incomplete",
        "yearStarted": 1995
      },
      "prisonNumber": "A1234AA",
      "setting": {
        "location": "Long Funk",
        "type": "custody"
      },
      "source": "rightfully"
    },
    {
      "addedBy": "Trace Nikolaus",
      "courseName": "Yellow Course",
      "createdAt": "Tue Nov 07 2023 05:09:08 GMT+0000 (Coordinated Universal Time)",
      "detail": "Confido charisma clamo consectetur.",
      "id": "eb357e5d-5416-43bf-a8d2-0dc8fd92162e",
      "outcome": {
        "status": "incomplete"
      },
      "prisonNumber": "A1234AA",
      "source": "usefully outside where"
    }
  ]
}
```

- The API fetches the generated JSON contract and plays the values in the [seed data sql file](https://github.com/ministryofjustice/hmpps-accredited-programmes-api/blob/550d2ec6d814272977fb4d02bddf21549a097bc6/src/test/resources/db/migration/R__test_data.sql#L13) against the contract generated above. Because none of those entities have a `null` `location` (what we call `setting` on the backend), the API tests fail, as the UI is apparently expecting the entity not to have the `setting` field on it.

yndajas: Hmm. Would generating data in our UI client tests that match the API test data not fix this problem then, or extending the test data where needed? I know we want to move away from the SQL file at some point anyway though

yndajas: It's interesting that it omits the `setting` rather than setting it as `null` though

PR #872, opened 2025-02-21
(APG-651b) Only ask sexual offence question for PNI Building Choices referral

jsrobertson:
## Context

As part of the new PNI drives referrals journey we no longer ask if they are in a female prison, because we can find out as we have the persons ID up front.

## Changes in this PR
- Update `PersonUtils.personFromPrisoner` to return `prisonId`
- Update` OrganisationUtils.organisationFromPrison` to return `female` boolean
- Update BuildingChoicesFormController to prefill womens prison question and make the input hidden.
  - I've kept the existing non hidden input there too for the time being if an editor happens to view the page from the programme directory.
- Add new integration test for Building Choices make a referral journey

## Screenshots of UI changes

image

## Release checklist

[Release process documentation](../blob/main/doc/how-to/perform-a-release.md)

### Pre-merge

### Post-merge


server/controllers/find/buildingChoicesFormController.ts
```diff
@@ -1,20 +1,46 @@
...
 
       FormUtils.setFieldErrors(req, res, ['isConvictedOfSexualOffence', 'isInAWomensPrison'])
-      FormUtils.setFormValues(req, res)
+      FormUtils.setFormValues(req, res, defaultValues)
 
       res.render('courses/buildingChoices/form/show', {
-        backLinkHref: isPniFind ? findPaths.pniFind.recommendedProgrammes({}) : findPaths.index({}),
-        pageHeading: "About the person you're referring",
+        backLinkHref,
+        hideWomensPrisonQuestion: Boolean(pniFindPrisonNumber && defaultValues.isInAWomensPrison),
```
thomaswilsonxyz, line 41: it feels odd to be treating `isInAWomensPrison` as a string, and not as a nullable boolean here - it took me a moment to understand what was happening here because:

```js
Boolean('false')
// true

Boolean('')
// false
```

Can we perhaps use 

```js
defaultValues.isInAWomensPrison: null | boolean = null
defaultValues.isInAWomensPrison = organisation?.female === 'true'

// ...later
if (defaultValues.isInAWomensPrison === null) ...
```

It took me bouncing around a few files to see why this had been set this way.

server/controllers/find/buildingChoicesFormController.ts
```diff
@@ -1,20 +1,46 @@
...
+      const pniFindPrisonNumber = req.session.pniFindAndReferData?.prisonNumber
+
+      if (pniFindPrisonNumber) {
+        const person = await this.personService.getPerson(req.user.username, pniFindPrisonNumber)
+        const organisation = person.prisonId
+          ? await this.organisationService.getOrganisation(req.user.token, person.prisonId)
+          : undefined
+
+        backLinkHref = findPaths.pniFind.recommendedProgrammes({})
+        defaultValues.isInAWomensPrison =
```
thomaswilsonxyz, line 31: Why have an object with one entry ?  I see we use it below in the `setFormValues`, but it feels a little over-abstract here, and we could

```ts
const isInAWomensPrison = organisation?.female === 'true'

FormUtils.setFormValues(req, res, { isInAWomensPrison })
```

without much difference

server/views/courses/buildingChoices/form/show.njk
```diff
@@ -22,23 +22,26 @@
         }) }}
       {% endif %}
 
-      <h1 class="govuk-heading-l">{{ pageHeading }}</h1>
-
-      <p class="govuk-body">Not all prisons can offer programmes to all prisoners. To see a list of suitable locations, tell us more about the person you're referring.</p>
+      {% set hintHtml %}
+      <p class="govuk-body">Not all prisons can offer programmes to all prisoners. To see a list of suitable locations, tell us whether the person youre referring has been convicted of a sexual offence.</p>
```
thomaswilsonxyz, line 26: `youre` -> `you're`

thomaswilsonxyz (commented):
This is quite a big PR, so I don't feel comfortable approving to myself just yet - especially as I have yet to get my service running locally so I can't verify

I've just left a few comments about my initial reception

server/controllers/find/buildingChoicesFormController.ts
jsrobertson: Ultimately `isInAWomensPrison` gets passed to the template where it is used by the [govukRadios component](https://design-system.service.gov.uk/components/radios/), which expects a `string` as the `value`.  To avoid the confusion between `'true'` and `true`, it could be an option to switch to `'yes'` (and `'no'`).

server/controllers/find/buildingChoicesFormController.ts
jsrobertson: Agreed on object with one entry so I've changed it but I have also ended up changing flow of the method so it returns early if a user is viewing the page and not on a PNI journey.  In this scenario we have to show the Womens prison question.  Happy to talk through it if needed.

thomaswilsonxyz (approved):

PR #910, opened 2025-03-18
(APG-694f) Integration tests for Transfer

jsrobertson:
## Context

Transferring to Building Choices functionality is mostly there and we needed some integration tests.

## Changes in this PR
Added tests for happy path and error scenarios.


Giruvagen (approved):
Approving with notes that we need an integration test for the generic error case as well as adding tests in for the CTA

PR #963, opened 2025-04-28
(APG-858) Remove use of /organisations/enabled endpoint

jsrobertson:
## Context

The /organisations/enabled endpoint was only ever meant to be short lived.

## Changes in this PR

- Switch to use existing `getAllOrganisations` method in reportsController.
- Delete any reference to `/organisations/enabled` endpoint

Note: We should remove the use of prison register API and instead use `/organisations/` and `/organisations/:code` from AcP API. However this is a bigger piece of work and outside the scope of this task.


Giruvagen (approved):

server/utils/organisationUtils.test.ts
```diff
@@ -76,12 +77,12 @@ describe('OrganisationUtils', () => {
...
+    describe('when a organisation has an `undefined` prisonId or prisonName', () => {
       it('filters out the organisation from the returned array', () => {
-        const enabledOrganisations: Array<EnabledOrganisation> = [
-          { code: 'B', description: 'B Prison' },
-          { code: undefined, description: 'A Prison' },
-          { code: 'A', description: undefined },
+        const enabledOrganisations: Array<Prison> = [
+          prisonFactory.build({ prisonId: 'B', prisonName: 'B Prison' }),
+          prisonFactory.build({ prisonId: undefined, prisonName: 'A Prison' }),
+          prisonFactory.build({ prisonId: 'A', prisonName: undefined }),
```
thomaswilsonxyz, line 85: i know it's just a re-move, but seems a little odd that prison with ID `A` has a name of `undefined`, and prison with name `A Prison` has an of `undefined` - seems to break the conventions around it

thomaswilsonxyz (commented):
Just chasing around some of the code, not one for now, can be in a follow-up commit, but I reckon we can delete the following:

- The static method `OrganisationUtils:organisationRadioItems`
- The `EnabledOrganisation` type

thomaswilsonxyz (approved):

thomaswilsonxyz (approved):
left a few comments, but lgtm

server/utils/organisationUtils.test.ts
jsrobertson: Agreed.

jsrobertson:
> * EnabledOrganisation

Have removed `OrganisationUtils.organisationRadioItems`.

`EnabledOrganisations` is auto generated and I suspect will be removed after the related API change has been merged and we re-generate.

server/controllers/reportsController.ts
```diff
@@ -60,10 +60,10 @@ export default class ReportsController {
             }),
           ),
         ),
-        this.referenceDataService.getEnabledOrganisations(req.user.username),
+        this.organisationService.getAllOrganisations(req.user.token),
```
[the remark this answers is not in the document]
jsrobertson: The service is now available to all organisations so the concept of enabling or disabling an organisation is now no longer necessary. This was required when we were in private beta and restricted the service to a limited number of organisations.

thomaswilsonxyz (approved):

PR #1064, opened 2026-04-16
Add Serious violent reoffending predictor to the OGRS4 version of the…

niaevans9:
Add Serious violent reoffending predictor to the OGRS4 version of the risks and needs page
Also a fix for the broken predictor scale (it was looking like this)

Screenshot 2026-04-16 at 13 03 34

image


rgforsyth (approved):

raby (commented):
LGTM

PR #1076, opened 2026-06-03
Update risk of serious recidivism terminology

niaevans9:
For all referrals, on the ‘Risks and Needs’ tab, under ‘Risks and alerts’ we have a ‘Risk of serious recidivism’ section.

The High/Medium/Low scores for this are currently against:

OSP/C

OSP/I

These acronyms need to be updated to:

OSP/DC

OSP/IIOC

### DIFF
--- a/server/@types/ui/index.d.ts
+++ b/server/@types/ui/index.d.ts
@@ -203,7 +203,7 @@ type OspBox = {
   dataTestId: string
   levelClass: string
   levelText: string
-  type: 'OSP/C' | 'OSP/I'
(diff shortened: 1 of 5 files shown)


server/@types/ui/index.d.ts
```diff
@@ -203,7 +203,7 @@ type OspBox = {
   dataTestId: string
   levelClass: string
   levelText: string
-  type: 'OSP/C' | 'OSP/I'
+  type: 'OSP/C' | 'OSP/DC' | 'OSP/I' | 'OSP/IIOC'
```
rgforsyth, line 206: Do we still need the old OSP/C & OSP/I acronyms?

rgforsyth (approved):

